PR #2, opened 2017-07-10
support Web App on Windows and FTP deployment

xscript:


maven-plugin-lib/src/main/java/com/microsoft/azure/maven/FTPUploader.java
```diff
@@ -0,0 +1,170 @@
...
+/**
+ * Utility class to upload directory to FTP server
+ */
+public class FTPUploader {
+    public static final String UPLOAD_START = "#%d Start uploading to FTP server: %s";
+    public static final String UPLOAD_SUCCESS = "Successfully upload files to FTP server: ";
+    public static final String UPLOAD_FAILURE = "Failed to upload files to FTP server after retries...";
+    public static final String UPLOAD_DIR_START = "Start uploading directory: %s --> %s";
+    public static final String UPLOAD_DIR_FINISH = "Finishing uploading directory: %s --> %s";
+    public static final String UPLOAD_DIR_FAILURE = "Failed to upload directory: %s --> %s";
```
ZhijunZhao: UPLOAD_DIR_FINISH -> UPLOAD_DIR_SUCCESS?

maven-plugin-lib/src/main/java/com/microsoft/azure/maven/FTPUploader.java
```diff
@@ -0,0 +1,170 @@
...
+     */
+    public FTPUploader(final Log log) {
+        this.log = log;
+        this.ftpClient = new FTPClient();
+    }
+
+    /**
+     * Upload directory to specified FTP server with retries.
+     *
+     * @param ftpServer
```
aovechki: is custom port number a valid scenario?

maven-plugin-lib/src/main/java/com/microsoft/azure/maven/FTPUploader.java
```diff
@@ -0,0 +1,170 @@
...
+        if (!isTargetDirectoryExist) {
+            ftpClient.makeDirectory(targetDirectoryPath);
+        }
+
+        final String nextLevelPrefix = logPrefix + "..";
+        for (File file : files) {
+            if (file.isFile()) {
+                uploadFile(file.getAbsolutePath(), targetDirectoryPath, nextLevelPrefix);
+            } else {
+                uploadDirectory(sourceDirectoryPath + "/" + file.getName(),
```
aovechki: is this cross-platform safe? Should we use Path class instead?

maven-plugin-lib/src/main/java/com/microsoft/azure/maven/FTPUploader.java
```diff
@@ -0,0 +1,170 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Utility class to upload directory to FTP server
+ */
+public class FTPUploader {
+    public static final String UPLOAD_START = "#%d Start uploading to FTP server: %s";
```
aovechki: Better: "Start" -> "Starting" or "Started" depending on the context

maven-plugin-lib/src/main/java/com/microsoft/azure/maven/FTPUploader.java
xscript: No. The FTP server is built into Web Apps. Custom port number is not supported.

maven-plugin-lib/src/main/java/com/microsoft/azure/maven/FTPUploader.java
xscript: Good suggestion.

webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/DeployFacadeBaseImpl.java
```diff
@@ -0,0 +1,87 @@
...
+    protected RuntimeHandler getRuntimeHandler() throws MojoExecutionException {
+        final JavaVersion javaVersion = getMojo().getJavaVersion();
+        final ContainerSetting containerSetting = getMojo().getContainerSettings();
+
+        if (javaVersion != null && containerSetting != null && !containerSetting.isEmpty()) {
+            throw new MojoExecutionException(RUNTIME_CONFIG_CONFLICT);
+        }
+
+        if (javaVersion != null) {
+            return new JavaRuntimeHandlerImpl(getMojo());
```
adashen: So if user specify javaVersion it will deploy the webapp windows?

xscript: Right.

maven-plugin-lib/src/main/java/com/microsoft/azure/maven/FTPUploader.java
xscript: I tried on both Windows and macOS. It works. But agree that we should use Path class to handle it more safely and elegantly.

xscript: Still use `/` for  `targetPath` because it is the path in remote FTP server. Using system `fileSeparator` will mess it up on Windows.

PR #5, opened 2017-07-13
fix missing war file

xscript:


xscript:
Find some code issue. Will submit a better fix later.

PR #7, opened 2017-07-19
throw exception when ftp reply code indicate failure

xscript:


maven-plugin-lib/src/main/java/com/microsoft/azure/maven/FTPUploader.java
```diff
@@ -157,15 +157,24 @@ protected void uploadFile(final FTPClient ftpClient, final String sourceFilePath
...
             ftpClient.changeWorkingDirectory(targetFilePath);
             ftpClient.storeFile(sourceFile.getName(), is);
-            log.info(String.format(UPLOAD_FILE_REPLY, logPrefix, ftpClient.getReplyCode(), ftpClient.getReplyString()));
-        } catch (Exception e) {
-            throw e;
-        } finally {
-            is.close();
+
+            final int replyCode = ftpClient.getReplyCode();
+            final String replyMessage = ftpClient.getReplyString();
```
ZhijunZhao: Put this line inside the if(){}?

PR #13, opened 2017-07-29
Function Maven Plugin prototype

xscript:
1. support function deployment using FTP;
2. implement `function.json` generation;
3. move common logic to `maven-plugin-lib`;


function-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
```diff
@@ -40,11 +52,19 @@ public String getRegion() {
...
     }
 
-    public List<Resource> getResources() {
-        return resources;
+    public FunctionApp getFunctionApp() {
+        FunctionApp app = null;
+        try {
+            app = getAzureClient().appServices().functionApps().getByResourceGroup(getResourceGroup(), getAppName());
+        } catch (Exception e) {
+            // Swallow exception
```
ZhijunZhao: My preference is to throw the exception.

xscript: @ZhijunZhao  This is to check the existence of the Function App. Azure Java SDK doesn't provide an API for such check.

PR #64, opened 2017-11-15
add batch mode for add function

jdneo:
This PR add the batch mode for maven add function goal.

When calling ```assureInputFromUser()``` for each property, we add a boolean param to the method to specify whether this property is required or not.

If the user is in batch mode and the property is required, the maven goal will throw MojoFailureException when the input in invalid, if is not required, we will use empty string to replace it.

If the user is in interactive mode, the experice is not changed as previous.

### DIFF
--- a/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AddMojo.java
+++ b/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AddMojo.java
@@ -168,11 +168,20 @@ public class AddMojo extends AbstractFunctionMojo {
         info("");
         info(FIND_TEMPLATE);
 
-        assureInputFromUser("template for new function",
-                getFunctionTemplate(),
-                getTemplateNames(templates),
-                this::setFunctionTemplate,
-                true);
+        if (settings != null && !settings.isInteractiveMode()) {
+            assureInputInBatchMode(getFunctionTemplate(),
+                    str -> getTemplateNames(templates)
+                            .stream()
+                            .filter(Objects::nonNull)
+                            .anyMatch(o -> o.equalsIgnoreCase(str)),
+                    this::setFunctionTemplate,
+                    true);
+        } else {
+            assureInputFromUser("template for new function",
+                    getFunctionTemplate(),
+                    getTemplateNames(templates),
+                    this::setFunctionTemplate);
+        }
 
         return findTemplateByName(templates, getFunctionTemplate());
     }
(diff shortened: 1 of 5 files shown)


xscript:
Upgrade the versions of `azure-functions-java-core` to `[1.0.0-beta-1,1.0.0)` at below places.
- https://github.com/Microsoft/azure-maven-plugins/blob/master/azure-functions-maven-plugin/pom.xml#L27
- https://github.com/Microsoft/azure-maven-plugins/blob/master/azure-functions-maven-plugin/src/it/http-trigger/pom.xml#L26

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AddMojo.java
```diff
@@ -378,6 +414,23 @@ protected void assureInputFromUser(final String prompt, final String initValue,
         }
     }
 
+    protected void assureInputInBatchMode(final String input, final Function<String, Boolean> validator,
+                                          final Consumer<String> setter, final boolean required)
+            throws MojoFailureException {
+        if (validator.apply(input)) {
+            info(FOUND_VALID_VALUE);
+            setter.accept(input);
+            return;
```
xscript: unnecessary return? or you can remove below else to avoid nesting if. Leave the decisions to u.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AddMojo.java
```diff
@@ -8,6 +8,8 @@
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.microsoft.azure.maven.function.template.FunctionTemplate;
+
```
xscript: Unnecessary blank line?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AddMojo.java
jdneo: Sure, removing below else to avoid nesting if.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AddMojo.java
```diff
@@ -378,6 +413,23 @@ protected void assureInputFromUser(final String prompt, final String initValue,
...
+                                          final Consumer<String> setter, final boolean required)
+            throws MojoFailureException {
+        if (validator.apply(input)) {
+            info(FOUND_VALID_VALUE);
+            setter.accept(input);
+            return;
+        }
+
+        if (required) {
+            throw new MojoFailureException(String.format("invalid input: %s", input));
```
xscript: `input` is probably `null`?

PR #70, opened 2017-12-13
add the selection for HttpTrigger's authLevel property

jdneo:
issue: #67


azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AddMojo.java
```diff
@@ -274,18 +276,38 @@ protected void preparePackageName() throws MojoFailureException {
         for (final String property : template.getMetadata().getUserPrompt()) {
             info(format("Trigger specific parameter [%s]", property));
 
+            final List<String> options = OptionTypedProperty.get(property) == null ?
```
adashen: how about just one method call like getOptionsforProperty?

jdneo: Good point

PR #71, opened 2017-12-25
not contain install-id and session-id in UA when allowTelemetry is false

jdneo:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
```diff
@@ -177,10 +177,15 @@ public String getPluginVersion() {
     }
 
     public String getUserAgent() {
-        return String.format("%s/%s %s:%s %s:%s",
-                getPluginName(), getPluginVersion(),
-                INSTALLATION_ID_KEY, getInstallationId(),
-                SESSION_ID_KEY, getSessionId());
+        if (isTelemetryAllowed()) {
```
xscript: I would prefer `return isTelemetryAllowed ? ... : ... ;`, but leave it up to u.

jdneo: Agree, that will be more concise, will change it later

PR #72, opened 2017-12-26
show privacy statement for first running

jdneo:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
```diff
@@ -327,6 +350,30 @@ protected void handleException(final Exception exception) throws MojoExecutionEx
         }
     }
 
+    private boolean isFirstRun(Properties prop) throws IOException {
```
xscript: Should we capture exceptions from this method to avoid blocking maven plugin execution?
Same concern applies to `updateConfigurationFile`.

PR #112, opened 2018-03-15
use <linuxRuntime> to set the linux runtime

jdneo:
According to the feedback. We should not overload the container setting but using a new setting tag which is parallel with `<javaVersion>` and `<containerSetting>`

So now:
`<javaVersion>` -> Web App on Windows
`<linuxRuntime>` -> Web App on Linux
`<containerSetting>` Web App for containers


ansyral:
javaVersion -> windowsJavaVersion ?

PR #116, opened 2018-03-20
Add unit tests and integration tests for Linux Web App

jdneo:
- Add unit test for LinuxRuntimeHandlerImpl
- Add integration test for linux webapp
- move getLinuxRunTimeStack() into utils and add unit test for it

### DIFF
--- a/azure-webapp-maven-plugin/src/it/linux-ftp/setup.groovy
+++ b/azure-webapp-maven-plugin/src/it/linux-ftp/setup.groovy
@@ -1,0 +1,17 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
+// Delete resources before test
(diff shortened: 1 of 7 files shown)


azure-webapp-maven-plugin/src/it/linux-ftp/setup.groovy
```diff
@@ -0,0 +1,17 @@
...
+ * license information.
+ */
+
+// Delete resources before test
+def clientId = System.getenv("CLIENT_ID")
+def tenantId = System.getenv("TENANT_ID")
+def key = System.getenv("KEY")
+def command = """
+    az login --service-principal -u ${clientId}  -p ${key} --tenant ${tenantId}
+    az group delete -y -n maven-webapp-it-rg-7
```
ansyral: what is the difference between this script and cleanup.groovy?

jdneo: setup is to sync-delete the resource group, clean up is async

Issue #90, opened 2018-02-22
java.lang.NoClassDefFoundError: org/springframework/context/ApplicationContext when building a function which calls Spring's Application Context


damadei:
### Plugin name and version
com.microsoft.azure.azure-functions-maven-plugin:0.1.10

### Plugin configuration in your `pom.xml`
			<plugin>
				<groupId>com.microsoft.azure</groupId>
				<artifactId>azure-functions-maven-plugin</artifactId>
				<configuration>
					<resourceGroup>java-functions-group</resourceGroup>
					<appName>${functionAppName}</appName>
					<region>${functionAppRegion}</region>
					<appSettings>
						<property>
							<name>FUNCTIONS_EXTENSION_VERSION</name>
							<value>beta</value>
						</property>
					</appSettings>
				</configuration>
				<executions>
					<execution>
						<id>package-functions</id>
						<goals>
							<goal>package</goal>
						</goals>
					</execution>
				</executions>
			</plugin>

### Expected behavior
Generate my maven package

### Actual behavior
I'm getting the error bellow only when my code is using applicationContext.getBean() method. If I use ApplicationContext class, it works.

So it works with this:

ApplicationContext appContext = new ClassPathXmlApplicationContext("classpath*:/application-context.xml");

If I add the following line, it stops working:

appContext.getBean("jdbcTemplate");

And the mvn clean package throws the following error:
Number of foreign imports: 1
import: Entry[import  from realm ClassRealm[maven.api, parent: null]]

-----------------------------------------------------

        at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:166)
        ... 21 more
Caused by: java.lang.NoClassDefFoundError: org/springframework/context/ApplicationContext
        at java.lang.Class.getDeclaredMethods0(Native Method)
        at java.lang.Class.privateGetDeclaredMethods(Class.java:2701)
        at java.lang.Class.getDeclaredMethod(Class.java:2128)
        at org.reflections.util.Utils.getMemberFromDescriptor(Utils.java:88)
        at org.reflections.util.Utils.getMethodsFromDescriptors(Utils.java:101)
        at org.reflections.Reflections.getMethodsAnnotatedWith(Reflections.java:482)
        at com.microsoft.azure.maven.function.handlers.AnnotationHandlerImpl.findFunctions(AnnotationHandlerImpl.java:43)
[...]

PR #119, opened 2018-03-26
function package error fix

jdneo:
fix: #90 

~~Solution: try to resolve the runtime dependencies from the output of the maven-dependency-plugin, and add them to the class path.~~

UPDATE
-----------------
sulution: get the compile-scope dependencies from the maven itself: adding `requiresDependencyResolution = ResolutionScope.COMPILE` in the `@Mojo` annotation.

### DIFF
--- a/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
+++ b/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
@@ -14,15 +14,20 @@ import com.microsoft.azure.maven.function.configurations.FunctionConfiguration;
 import com.microsoft.azure.maven.function.handlers.AnnotationHandler;
 import com.microsoft.azure.maven.function.handlers.AnnotationHandlerImpl;
 
+import org.apache.maven.model.Plugin;
+import org.apache.maven.model.PluginExecution;
 import org.apache.maven.model.Resource;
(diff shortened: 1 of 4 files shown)


azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
```diff
@@ -108,6 +117,44 @@ protected URL getTargetClassUrl() throws Exception {
...
+    /**
+     * This method will search the output directory which configured in the Maven Dependency Plugin.
+     * Maven Dependency Plugin will extract the runtime jars into the output directory,
+     * and these jars will be used for annotation search using reflection.
+     * @return URLs for the runtime needed jars
+     */
+    protected List<URL> getDependencyArtifactUrls() {
+        final List<URL> urlList = new ArrayList<>();
+        final Plugin dependencyPlugin = this.getProject().getPlugin(DEPENDENCY_PLUGIN_KEY);
+        if (dependencyPlugin != null) {
```
ansyral: so it requires the pom.xml to have "copy-dependencies" goal? is it a walk around?

jdneo: Emm, that is a work around. But I just find a better solution for that.

Code is updated, sorry for the confusion.

PR #123, opened 2018-04-05
support create web app to an existing app service plan

### LINKED ISSUE #62: Azure Maven Plugin for Web Apps Does NOT Allow the Use of Existing App Service Plans
### Plugin name and version
Azure Maven Plugin for Web Apps 0.1.5

### Plugin configuration in your `pom.xml`
Any configuration

### Expected behavior
Should allow the use of existing app service plans

### Actual behavior
Does not allow the use of existing app service plans. It creates an app service plan for every deploy. This limits the use of the Maven plugin because there are subscription limits - https://docs.microsoft.com/en-us/azure/azure-subscription-service-limits#app-service-limits. There are known issues to cleanup app service plans which further aggravates developers. We have run into these limits several times now and then had to open support tickets to clean up app service plans. This creates FRICTION for developers.

The Azure Management Libraries for Java make it simple to use existing App Service Plans.

### Steps to reproduce the problem
Just keep running one of the sample deployments to Azure Web Apps

jdneo:
resolve #62 

The pom configuration will be:

```xml
...
<!-- Web App information -->
<resourceGroup>app-group</resourceGroup>
<appName>app-name</appName>

<!-- App Service Plan information -->
<appServicePlanResourceGroup>plan-group</appServicePlanResourceGroup>
<appServicePlanName>plan-name</appServicePlanName>       
...
```

### DIFF
--- a/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/WebAppUtils.java
+++ b/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/WebAppUtils.java
@@ -58,14 +58,14 @@ public class WebAppUtils {
     public static WithDockerContainerImage defineLinuxApp(final AbstractWebAppMojo mojo, final AppServicePlan plan)
             throws Exception {
         final String resourceGroup = mojo.getResourceGroup();
-        final boolean isResourceGroupExist = mojo.getAzureClient().resourceGroups().contain(resourceGroup);
 
         assureLinuxPlan(plan);
 
         final ExistingLinuxPlanWithGroup existingLinuxPlanWithGroup = mojo.getAzureClient().webApps()
                 .define(mojo.getAppName())
                 .withExistingLinuxPlan(plan);
-        return isResourceGroupExist ? existingLinuxPlanWithGroup.withExistingResourceGroup(resourceGroup) :
+        return mojo.getAzureClient().resourceGroups().contain(resourceGroup) ?
+                existingLinuxPlanWithGroup.withExistingResourceGroup(resourceGroup) :
                 existingLinuxPlanWithGroup.withNewResourceGroup(resourceGroup);
     }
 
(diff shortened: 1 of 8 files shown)


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/WebAppUtils.java
```diff
@@ -42,14 +55,88 @@ public static void assureWindowsWebApp(final WebApp app) throws MojoExecutionExc
         }
     }
 
-    public static WithNewAppServicePlan defineApp(final AbstractWebAppMojo mojo) throws Exception {
-        final WithGroup<WithNewAppServicePlan> withGroup = mojo.getAzureClient().webApps()
+    public static WithDockerContainerImage defineLinuxApp(final AbstractWebAppMojo mojo, final AppServicePlan plan)
+            throws Exception {
+        final String resourceGroup = mojo.getResourceGroup();
```
ansyral: move the variable `resourceGroup` closer to where it gets used?

ansyral: same to `defineWindowsApp`

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/WebAppUtils.java
```diff
@@ -42,14 +55,88 @@ public static void assureWindowsWebApp(final WebApp app) throws MojoExecutionExc
...
+    }
+
+    private static void assureLinuxPlan(final AppServicePlan plan) throws MojoExecutionException {
+        if (!plan.operatingSystem().equals(OperatingSystem.LINUX)) {
+            throw new MojoExecutionException(String.format(SERVICE_PLAN_NOT_APPLICABLE,
+                    plan.name(), OperatingSystem.LINUX.name()));
+        }
+    }
+
+    public static WithCreate defineWindowsApp(final AbstractWebAppMojo mojo, final AppServicePlan plan)
```
ansyral: seems the logic is same except for `assurePlan`? is it possible to extract the same logic to one method?

jdneo: Emm, besides the `assurePlan()`, their return types are also different. Any suggestions?

ansyral: could the return type be a base type?

jdneo: Seems could not.

- WithDockerContainerImage: https://github.com/Azure/azure-libraries-for-java/blob/master/azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/WebApp.java#L240
- WithCreate: https://github.com/Azure/azure-libraries-for-java/blob/master/azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/WebApp.java#L302

ansyral: i see...

PR #142, opened 2018-04-23
Support war deploy in web app plugin

jdneo:


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/WarArtifactHandlerImpl.java
```diff
@@ -0,0 +1,80 @@
...
+    public static final int DEFAULT_MAX_RETRY_TIMES = 3;
+    public static final String UPLOAD_FAILURE = "Failed to deploy the war file to server, " +
+            "retrying immediately (%d/%d)";
+
+    public WarArtifactHandlerImpl(final AbstractWebAppMojo mojo) {
+        this.mojo = mojo;
+    }
+
+    @Override
+    public void publish(final List<Resource> resources) throws Exception {
```
ansyral: resources not used for War?

jdneo: Yes. How about make resources as a member in which class needs it?

jdneo: Updated, get the resources from `mojo` and `mojo` is already a member of the Hander

PR #161, opened 2018-05-28
refine integration tests

jdneo:
resolve: #155


azure-functions-maven-plugin/src/it/http-trigger-with-3rd-lib/pom.xml
```diff
@@ -1,140 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
ansyral: why delete the project?

jdneo: Because this project is using shaded plugin to generate a fat jar. Since the service side has already support non-fat jar functions and fat-jar is not a recommended way to package Function project, so I decide to remove this.

ansyral: then it isn't the reason why integration tests fail?

azure-functions-maven-plugin/src/it/http-trigger/cleanup.groovy
```diff
@@ -5,14 +5,24 @@
...
 
 // Verify Azure Functions
-def url = "https://maven-functions-it-${timestamp}-1.azurewebsites.net/api/hello?name=Azure".toURL()
-try {
-    url.getText() // warm up
-} catch (Exception e) {
-    // ignore warm-up exception
+def url = "https://maven-functions-it-${timestamp}-1.azurewebsites.net/api/hello?json={\"body\":\"Azure\"}".toURL()
+
+// Functions need some time to warm up
```
ansyral: why not just wait for some time?

ansyral: try 5 times, en, I might think it is because Azure function isn't stable or has some network issue.

azure-functions-maven-plugin/src/it/http-trigger-with-3rd-lib/pom.xml
jdneo: Yes, the removal is just a refine.

azure-functions-maven-plugin/src/it/http-trigger/cleanup.groovy
jdneo: We need to trigger the function to wake it up. Only waiting for a while is not enough. 

Emm, yes 5 times is to make sure the integration test is not affected by stability issues, but if it still fails after 5 times, then something must go wrong.

jdneo: Or do you think 5 times are too much?

ansyral: No, what I mean here is trying multiple times until the contents echo back might hide other issues, but since you mention that we can only wake up the function by manually triggering it and it takes some time to wake up, then it is fine.

PR #164, opened 2018-05-30
Better run instructions in case issue happens

brunoborges:
Signed-off-by: Bruno Borges <bruno.borges@microsoft.com>


brunoborges:
@jdneo can you take a look?

jeffhollan:
Nice! Thanks for the contribution @brunoborges

PR #166, opened 2018-05-31
Update README.md

brunoborges:
Be more clear with the service brand, the plugin version, and use pluginManagement.


brunoborges:
@selvasingh FYI

brunoborges:
@jdneo mind to check?

jdneo:
@brunoborges Thanks. I was OOF recently so sorry for the late reply.

The changes look good to me. I changes the base from `master` to `develop`.

PR #167, opened 2018-06-01
AppService: Deployment of uber JAR files (like Spring Boot)

brunoborges:
This PR has two major features: #168 

1. Deployment of JAR files to Web App (Windows) with a `web.config` generated on the fly.
2. `deploymentType` no longer required to set. When not set, plugin will look at `packaging` to decide (usually, it is WAR or JAR).


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/DeployMojo.java
```diff
@@ -9,6 +9,7 @@
 import com.microsoft.azure.management.appservice.WebApp;
 import com.microsoft.azure.management.appservice.WebApp.DefinitionStages.WithCreate;
 import com.microsoft.azure.management.appservice.WebApp.Update;
+import com.microsoft.azure.maven.webapp.configuration.DeploymentType;
```
jdneo: Unused import?

azure-webapp-maven-plugin/pom.xml
```diff
@@ -211,7 +211,7 @@
                 <directory>src/main/resources</directory>
                 <filtering>true</filtering>
                 <includes>
-                    <include>**/ApplicationInsights.xml</include>
+                    <include>**/*</include>
```
jdneo: Why we need to change to `*` here?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/JarArtifactHandlerImpl.java
```diff
@@ -0,0 +1,103 @@
...
+
+        // FTP Upload all to wwwroot
+        uploadDirectoryToFTP();
+    }
+
+    private String getJarCommand() {
+        return mojo.getJarCommand() == null ? DEFAULT_JAR_COMMAND : mojo.getJarCommand();
+    }
+
+    private void generateWebConfigFile(String jarFileName) throws Exception {
```
jdneo: Emm, we have several IO operations in this method. Instead of just throw the Exceptions, I prefer we catch them and print some logs here, then we can re-throw them again.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/configuration/DeploymentType.java
```diff
@@ -3,18 +3,46 @@
...
+
+    /**
+     * Identifies the proper artifact handler based on the type of the deployment configured in the Maven Mojo object.
+     * It will inspect first the <deploymentType> property. If NONE, it will look into <packaging> of the Maven
+     * artifact.
+     *
+     * @param mojo for the Maven project
+     * @return an ArtifactHandler mapped to the deployment type identified
+     */
+    public ArtifactHandler fromMojo(AbstractWebAppMojo mojo) {
```
jdneo: change method name to `getArtifactFromMojo`?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/configuration/DeploymentType.java
```diff
@@ -26,8 +54,53 @@ public static DeploymentType fromString(final String input) {
...
+
+    static class NONEHandler implements Function {
+        public ArtifactHandler apply(AbstractWebAppMojo m) {
+            switch (m.getProject().getPackaging()) {
+                case "war":
+                    return new WarArtifactHandlerImpl(m);
+                case "jar":
+                    return new JarArtifactHandlerImpl(m);
+                default:
+                    throw new RuntimeException("You must set a packaging type of (jar,war) or a " +
```
jdneo: nit: add a space after `,` in `(jar,war)`

azure-webapp-maven-plugin/pom.xml
brunoborges: There is a web.config.template file

brunoborges:
@jdneo hey Sheng, thanks for the review. I pushed the changes. Pls take a look again.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/HandlerFactoryImpl.java
```diff
@@ -67,24 +67,14 @@ public SettingsHandler getSettingsHandler(final AbstractWebAppMojo mojo) {
...
-            case UNKNOWN:
-                throw new MojoExecutionException(UNKNOWN_DEPLOYMENT_TYPE);
-            case WAR:
-                return new WarArtifactHandlerImpl(mojo);
-            case FTP:
-            default:
-                return new FTPArtifactHandlerImpl(mojo);
-        }
+    public ArtifactHandler getArtifactHandler(final AbstractWebAppMojo mojo) {
+        return mojo.getDeploymentType().fromMojo(mojo);
```
jdneo: Here we also need to change to `getArtifactHandlerFromMojo()`

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/configuration/DeploymentType.java
```diff
@@ -26,8 +54,53 @@ public static DeploymentType fromString(final String input) {
...
+
+    static class NONEHandler implements Function {
+        public ArtifactHandler apply(AbstractWebAppMojo m) {
+            switch (m.getProject().getPackaging()) {
+                case "war":
+                    return new WarArtifactHandlerImpl(m);
+                case "jar":
+                    return new JarArtifactHandlerImpl(m);
+                default:
+                    throw new RuntimeException("You must set a packaging type of (jar, war) or a " +
```
jdneo: I prefer using `MojoExecutionException` here.

jdneo:
@brunoborges Cool, thank you for updating. Some more comments. 😄

brunoborges:
Should be good now @jdneo  :)

jdneo:
@brunoborges  Yes it looks fine to me. Just one place about the `<include>**/*</include>` could you also take a look?

About the schedule: since the service side will add Java SE support for Linux at the end of this week. We are planning to release a new version of Maven Web App plugin at early next week to support Java SE image.

Then we will look back to the Jar deploy, which may contain several further works:
- Merge this PR
- Re-organize the deploy logic, to make sure war, jar & ftp can works for all kinds of Web Apps (Windows, Linux, Containers)
- Add unit test
- Add integration test
- Update documents & README
- Run tests & verification

The ETA for adding Jar deploy will before the end of June. Is this looks good to you?

brunoborges:
Yes, end of June looks fine! Until then I can use a local version for my demos.

Regarding `<include>**/*</include>`, we can be specific to which files (app insights config, and the web.config.template), but I see no issue with the 'include all' approach, since this is our own code and we know which files will be included.

azure-webapp-maven-plugin/pom.xml
brunoborges: Fixed in 9d8d403

jdneo:
@brunoborges I made some changes. Please take a look

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/configuration/DeploymentType.java
```diff
@@ -63,11 +63,11 @@ public static DeploymentType fromString(final String input) {
     }
 
     // TODO: Change to lambda once on Java 8+
-    interface Function<T, R> {
+    interface Handler {
```
brunoborges: The reason this interface was called Function<T, R> was because it would be simpler to refactor to Java 8 later. But It is no big deal anyways.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/AbstractWebAppMojo.java
```diff
@@ -235,15 +235,15 @@
...
      *
-     * @since 1.2.0
+     * @since 1.3.0
      */
     @Parameter(property = "webapp.jarFile")
     protected String jarFile;
 
      /**
-     * Jar command arguments to be used by web.config
+     * Jar command arguments to be used by web.config (Web App on Windows)
```
brunoborges: We need to work on this feature and make this OS agnostic. <jarCommand> should be supported for JAR deployment, whether that's on Linux or Windows.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/JarArtifactHandlerImpl.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public static final String FILE_IS_NOT_JAR = "The deployment file is not a jar typed file.";
+    public static final String FIND_JAR_FILE_FAIL = "Failed to find the jar file: '%s'";
+
+
+    public static final String DEFAULT_LINUX_JAR_NAME = "app.jar";
+    public static final String JAR_CMD = "__JAR_COMMAND__";
+    public static final String DEFAULT_JAR_COMMAND = "-Djava.net.preferIPv4Stack=true "
+            + "-Dserver.port=%HTTP_PLATFORM_PORT% "
+            + "-jar &quot;%HOME%\\\\site\\\\wwwroot\\\\FILENAME&quot;";
```
Eskibear: `FILENAME`

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/JarArtifactHandlerImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        try (final InputStream is = getClass().getResourceAsStream("web.config.template")) {
+            templateContent = IOUtils.toString(is, "UTF-8");
+        } catch (IOException e) {
+            mojo.getLog().error(READ_WEB_CONFIG_TEMPLATE_FAIL);
+            throw e;
+        }
+
+        final String webConfigFile = templateContent
+                .replaceAll(JAR_CMD, getJarCommand())
+                .replaceAll("FILENAME", jarFileName);
```
Eskibear: `FILENAME`

brunoborges: Good call. FILENAME should only be replaced when using default command.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/AbstractWebAppMojo.java
jdneo: Yes but currently the executable jar support for Web App is different between Windows and Linux at server side.

web.config is a hack for Windows Web App. Linux Web App do not have such kind of file.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/configuration/DeploymentType.java
jdneo: I see.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/JarArtifactHandlerImpl.java
jdneo: Addressed

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/configuration/DeploymentType.java
```diff
@@ -26,8 +55,53 @@ public static DeploymentType fromString(final String input) {
...
+
+    static class NONEHandler implements Handler {
+        public ArtifactHandler apply(AbstractWebAppMojo m) throws MojoExecutionException {
+            switch (m.getProject().getPackaging()) {
+                case "war":
+                    return new WarArtifactHandlerImpl(m);
+                case "jar":
+                    return new JarArtifactHandlerImpl(m);
+                default:
+                    throw new MojoExecutionException("You must set a packaging type of (jar, war) or a " +
```
SummerSun: further check the deployment type before throwing exception

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/HandlerFactoryImpl.java
```diff
@@ -14,13 +14,11 @@
 import org.apache.maven.plugin.MojoExecutionException;
 
 public class HandlerFactoryImpl extends HandlerFactory {
-    public static final String RUNTIME_CONFIG_CONFLICT = "Conflict settings found. <javaVersion>, <linuxRuntime>" +
-            "and <containerSettings> should not be set at the same time.";
-    public static final String NO_RUNTIME_HANDLER = "Not able to process the runtime stack configuration; " +
-            "please check <javaVersion>, <linuxRuntime> or <containerSettings> tag in pom.xml";
+    public static final String RUNTIME_CONFIG_CONFLICT = "Conflict settings found. <javaVersion>, <linuxRuntime>"
```
SummerSun: do we have to throw the conflict exception, or, just pick/check the right configuration according to the deployment type?

jdneo: runtime is for the Linux, Windows or Containers. It's hard to infer the runtime according to the deployment type user is using.

jdneo:
@brunoborges Agree that we need to make the setting to be OS agnostic. I think the problem is because there is a gap for jar file support between Linux Web App and Windows Web App.

Windows Web App needs a hack: using web.config.

While for Linux Web App, what user needs to do is just upload a jar file named "app.jar", then server side will load it automatically. If user wants to customize his Java property, he can add `JAVA_OPTS` to application settings to achieve that. If he wants to do more customization, he can upload a startup script, set `INIT_SCRIPT` pointing to that file. That looks more suitable for FTP deployment to me. 

For now, can we just remove the jarCommand setting in the plugin? We just use the default command, if any user request such kind of feature, we can revisit it and see how can we add it to the configuration.

Does that make sense?

brunoborges:
It does. Mind to push the change?

jdneo:
Sure, I'll update it.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/HandlerFactoryImpl.java
brunoborges: @jdneo I think we should simplify this. Instead of <linuxRuntime>, we should have <environment>, and then choices being: windows, linux, container.

jdneo: @brunoborges Agree, Let’s put this in #200. We can focus on the configuration in the next release after jar deploy.

jdneo:
@brunoborges @SummerSun @Eskibear Please help review the newest codes and let me know if anything needs update. 😃

brunoborges:
@Eskibear mind to review/approve again? I ended up merging @jdneo's changes and pushed again to my repo (which triggered a PR review)

Issue #170, opened 2018-06-04
Updating package & artifact name


jdneo:
The service team has rename the dependency: `azure-functions-java-core` to `azure-functions-java-library`

We need to update related packages to align with the new updates.

PR #171, opened 2018-06-04
Update java function dependency

jdneo:
Fix: #170 

Three changes here: 
- `com.microsoft.azure.serverless.functions` to `com.microsoft.azure.functions`
- `azure-functions-java-core` to `azure-functions-java-library`
- Remove APIHub bindings since it's not supported in V2 runtime


azure-functions-maven-plugin/src/it/http-trigger/src/main/java/com/microsoft/azure/Function.java
```diff
@@ -7,8 +7,8 @@
 package com.microsoft.azure;
 
 import java.util.*;
-import com.microsoft.azure.serverless.functions.annotation.*;
-import com.microsoft.azure.serverless.functions.*;
+import com.microsoft.azure.functions.annotation.*;
+import com.microsoft.azure.functions.*;
```
pragnagopa: Please rename function from hello to HttpTriggerJava similar to  [HttpTriggerCSharp.cs#L25](https://github.com/Azure/azure-functions-templates/blob/dev/Functions.Templates/Templates/HttpTrigger-CSharp/HttpTriggerCSharp.cs#L25)

jdneo: @pragnagopa @JunyiYi Done.

jdneo:
@pragnagopa The PR for templates repo has been sent out: https://github.com/Azure/azure-functions-templates/pull/695

PR #215, opened 2018-07-04
share common code between integration test pre/post build script

SummerSun:


azure-webapp-maven-plugin/src/test/java/com/microsoft/azure/maven/webapp/utils/TestUtils.java
```diff
@@ -0,0 +1,34 @@
...
+        final String azLoginTemplate = "az login --service-principal -u %s -p %s --tenant %s";
+        final String azDelteTemplate = "az group delete -y -n %s%s";
+
+        final String[] commands = { String.format(azLoginTemplate, clientId, key, tenantId),
+                String.format(azDelteTemplate, resourceGroupName, waitForOperationFinish ? "" : " --no-wait"),
+                "az logout" };
+
+        if (System.getProperty("os.name").contains("Windows")) {
+            for (final String command : commands) {
+                Runtime.getRuntime().exec(String.format("cmd /c %s", command)).waitFor();
```
akaroml: Does waitFor mean the thread will be paused until the command finishes?
We need to be conscious of any blocking behavior.

jdneo: @akaroml yes it is. This is to make sure all the cleanup tasks can be addressed before an integration task begins.

SummerSun: @akaroml azure cli supports delete azure resource group without waiting for it to finish and we use that command in cleanup script, but not the setup.

PR #216, opened 2018-07-05
Add AUTO deployment type and defaults to it

brunoborges:
Signed-off-by: Bruno Borges <bruno.borges@microsoft.com>


jdneo:
@brunoborges  thanks for the PR, LGTM. 

Just pushed a commit to fix findgbus issues

PR #234, opened 2018-07-10
support web app deployment slot -part2

SummerSun:
* Implement IDeployTargetAdapter interface with DeploymentSlot

* Decide the deploy target from pom configuration

* Webapp must exists when deploy to slot, create new slot if not exits

* Update existing unit tests

* DeploymentSlotHandlerImpl unit tests, integrations tests will be included in next change list


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/DeploymentSlotHandlerImpl.java
```diff
@@ -0,0 +1,80 @@
...
+        this.mojo.getLog().info(CREATE_DEPLOYMENT_SLOT);
+
+        createDeploymentSlotWithConfigurationSource(app, slotName, slotSetting.getConfigurationSource());
+
+        this.mojo.getLog().info(CREATE_DEPLOYMENT_SLOT_DONE);
+    }
+
+    protected void createDeploymentSlotWithConfigurationSource (final WebApp app, final String slotName,
+                                                              final String configurationSource) {
+        final DeploymentSlot cSlot = this.mojo.getDeploymentSlot(app, configurationSource);
```
akaroml: what does cSlot mean?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/DeploymentSlotHandler.java
```diff
@@ -0,0 +1,14 @@
...
+ * license information.
+ */
+
+package com.microsoft.azure.maven.webapp.handlers;
+
+import com.microsoft.azure.maven.auth.AzureAuthFailureException;
+import org.apache.maven.plugin.MojoExecutionException;
+
+public interface DeploymentSlotHandler {
+    void handleDeploymentSlot() throws AzureAuthFailureException, MojoExecutionException;
```
akaroml: what does "handle" indicate?
in the implementation, i only see create if not exists operation. please be explicit if there are no other possibilities.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/DeploymentSlotHandlerImpl.java
```diff
@@ -0,0 +1,80 @@
...
+package com.microsoft.azure.maven.webapp.handlers;
+
+import com.microsoft.azure.management.appservice.DeploymentSlot;
+import com.microsoft.azure.management.appservice.WebApp;
+import com.microsoft.azure.maven.auth.AzureAuthFailureException;
+import com.microsoft.azure.maven.webapp.AbstractWebAppMojo;
+import com.microsoft.azure.maven.webapp.configuration.DeploymentSlotSetting;
+import org.apache.maven.plugin.MojoExecutionException;
+
+public class DeploymentSlotHandlerImpl implements DeploymentSlotHandler {
```
akaroml: will there be other handlers other than this?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/DeploymentSlotHandlerImpl.java
SummerSun: will rename to ```copyConfigurationFromSlot```

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/DeploymentSlotHandlerImpl.java
SummerSun: currently no, and this makes it more like a Util class.

SummerSun: Rethink about it, make DeploymentSlotUtils to handle deployment get/check/create functionalities.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/AbstractWebAppMojo.java
```diff
@@ -348,6 +363,21 @@ public WebApp getWebApp() throws AzureAuthFailureException {
         return null;
     }
 
+    public DeploymentSlot getDeploymentSlot(final WebApp app, final String slotName) {
+        DeploymentSlot slot = null;
+        if (!StringUtils.isEmpty(slotName)) {
```
jdneo: Does StringUtils has `isNotEmpty()` method?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/AbstractWebAppMojo.java
```diff
@@ -348,6 +363,21 @@ public WebApp getWebApp() throws AzureAuthFailureException {
         return null;
     }
 
+    public DeploymentSlot getDeploymentSlot(final WebApp app, final String slotName) {
+        DeploymentSlot slot = null;
+        if (!StringUtils.isEmpty(slotName)) {
+            try {
+                slot = app.deploymentSlots().getByName(slotName);
+            } catch (NoSuchElementException deploymentSlotNotExistException) {
```
jdneo: swallow the deploymentSlotNotExistException is fine here. But we need to differentiate:
- user doesn't set these configuration
- user set but the slot name is invalid

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/configuration/DeploymentSlotSetting.java
```diff
@@ -0,0 +1,31 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
+package com.microsoft.azure.maven.webapp.configuration;
+
+/**
+ * Deployment SLot setting class.
```
jdneo: SLot -> Slot

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/AbstractWebAppMojo.java
SummerSun: Actually there is ^_^

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/DeploymentSlotHandler.java
```diff
@@ -0,0 +1,78 @@
...
+public class DeploymentSlotHandler {
+    private static final String INVALID_SLOT_SETTINGS = "<deploymentSlotSetting> is NULL." +
+            "Need configure it in pom.xml for deploying it to deployment slot.";
+    private static final String EMPTY_CONFIGURATION_SOURCE =
+            "Empty configuration source, create a brand new deployment slot without any configuration.";
+    private static final String DEFAULT_CONFIGURATION_SOURCE =
+            "Null or unrecognized configuration source, create deployment slot and copy configuration from parent.";
+    private static final String CREATE_DEPLOYMENT_SLOT = "Start create deployment slot...";
+    private static final String CREATE_DEPLOYMENT_SLOT_DONE = "Successfully created deployment slot.";
+    private static final String INVALID_SLOT_NAME = "<slotName> is invalid for creating deployment slot.";
```
jdneo: Can be more specific here. Here `invalid` actually means empty, right?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/DeploymentSlotHandler.java
```diff
@@ -0,0 +1,78 @@
...
+        this.mojo = mojo;
+    }
+
+    public void createDeploymentSlotIfNotExist() throws MojoExecutionException, AzureAuthFailureException {
+        final DeploymentSlotSetting slotSetting = this.mojo.getDeploymentSlotSetting();
+        assureValidSlotSetting(slotSetting);
+
+        final WebApp app = this.mojo.getWebApp();
+        final String slotName = slotSetting.getSlotName();
+        final DeploymentSlot slot = this.mojo.getDeploymentSlot(app, slotName);
```
jdneo: if `slot` is only used in the if block, then we can remove this temp variable for now.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/DeploymentSlotHandler.java
```diff
@@ -0,0 +1,78 @@
...
+                                        final String configurationSource) throws MojoExecutionException {
+        assureValidSlotName(slotName);
+        final DeploymentSlot configurationSourceSlot = this.mojo.getDeploymentSlot(app, configurationSource);
+        final DeploymentSlot.DefinitionStages.Blank definedSlot = app.deploymentSlots().define(slotName);
+
+        this.mojo.getLog().info(CREATE_DEPLOYMENT_SLOT);
+
+        if (configurationSourceSlot != null) {
+            definedSlot.withConfigurationFromDeploymentSlot(configurationSourceSlot).create();
+        } else if ("".equalsIgnoreCase(configurationSource)) {
```
jdneo: `""` is also not null, so this block will not be executed actually. The if block here is buggy.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/configuration/ConfigurationSourceType.java
```diff
@@ -0,0 +1,31 @@
...
+ * license information.
+ */
+
+package com.microsoft.azure.maven.webapp.configuration;
+
+import org.codehaus.plexus.util.StringUtils;
+
+import java.util.Locale;
+
+public enum ConfigurationSourceType {
```
jdneo: Let's put some comments here to describe what's the meaning of each Enum

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/DeploymentSlotHandler.java
```diff
@@ -0,0 +1,101 @@
...
+    protected void assureValidSlotSetting(final DeploymentSlotSetting slotSetting) throws MojoExecutionException {
+        if (slotSetting == null) {
+            throw new MojoExecutionException(INVALID_SLOT_SETTINGS);
+        }
+    }
+
+    protected void assureValidSlotName(final String slotName) throws MojoExecutionException {
+        final Pattern pattern = Pattern.compile(SLOT_NAME_PATTERN, Pattern.CASE_INSENSITIVE);
+
+        if (!pattern.matcher(slotName).matches()) {
```
jdneo: Check NPE?

PR #246, opened 2018-07-13
Readme update for new deployment type

jdneo:
@Matthew-Dong @brunoborges Please help review the readme update when you have time. Thanks.


jdneo:
@brunoborges Do you mean we need to explicitly show Jar and War deployment type?

For now, there's a sentence says: 

> Note: If you do not want the plugin to inspect the `<packaging>` field. Just explicitly set the `<deploymentType>` to `jar` or `war`.

brunoborges:
I think we could be more explicit with the available options. And document
each independently (and AUTO after covering WAR and JAR, so reader
understands easily).

jdneo:
I see. So the order is:
1. WAR
2. JAR
3. AUTO
4. FTP
5. NONE

brunoborges:
Yeah, I like that!

jdneo:
@brunoborges updated.

PR #248, opened 2018-07-13
add deployment slot guidance in README

SummerSun:


azure-webapp-maven-plugin/README.md
```diff
@@ -275,6 +275,30 @@ Property | Description
...
+Property | Required | Description
+---|---|---
+`<slotName>` | true | Specifies the slot name. Valid slot name must match this kind of regex expression: [a-zA-Z0-9-]{1, 60}. 
+`<configurationSource>` | false | The configuration source used to create a deployment slot.
+
+> Note`<slotName>` is pure slot name, not the name with webapp name as prefix in the portal.
+
+If you specify an existing deployment slot to do the deploy, `<configurationSource>` will be ignored even it's configured.
+
+Or, if you want to do the deployment slot create and deploy at the same time, you could specify configuration source as you like.
```
Matthew-Dong: Suggest changing to -> "Or, if you want to create a new deployment slot create and deploy to it at the same time, you could specify configuration source as you like."

azure-webapp-maven-plugin/README.md
```diff
@@ -275,6 +275,30 @@ Property | Description
...
+
+`<configurationSource>` could be set to **New** to create a brand new slot without any configuration,
+or set to **Parent** to create a slot and copy the configuration from parent webapp.
+Besides, you could set it another existing deployment slot name to copy configuration from during creation.
```
Matthew-Dong: Suggest changing to -> "Besides, you could set it to the name of another existing deployment slot. It will copy the configuration of that slot during creation."

SummerSun: Taken

Issue #233, opened 2018-07-10
Incorrect message after webapp war deploy failed after retry 3 times


SummerSun:
![image](https://user-images.githubusercontent.com/5179063/42493710-a393c152-8450-11e8-84bc-25c64bb34cab.png)

Build end with SUCCESS but it actually failed after 3 times try.

PR #255, opened 2018-07-17
report deploy failure when war deploy failed after three times trying

SummerSun:
fix issue #233

### DIFF
--- a/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/WarArtifactHandlerImpl.java
+++ b/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/WarArtifactHandlerImpl.java
@@ -19,7 +19,7 @@ public class WarArtifactHandlerImpl implements ArtifactHandler  {
 
     public static final String FILE_IS_NOT_WAR = "The deployment file is not a war typed file.";
     public static final String FIND_WAR_FILE_FAIL = "Failed to find the war file: '%s'";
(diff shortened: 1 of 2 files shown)


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/WarArtifactHandlerImpl.java
```diff
@@ -39,15 +39,19 @@ public void publish(IDeployTargetAdapter deployTarget) throws MojoExecutionExcep
...
             try {
                 deployTarget.warDeploy(war, path);
                 return;
             } catch (Exception e) {
-                mojo.getLog().warn(String.format(UPLOAD_FAILURE, retryCount, DEFAULT_MAX_RETRY_TIMES));
+                mojo.getLog().warn(String.format(UPLOAD_FAILURE, e.getMessage(), retryCount, DEFAULT_MAX_RETRY_TIMES));
             }
         }
+
+        if (retryCount > DEFAULT_MAX_RETRY_TIMES) {
```
jdneo: Can we add this cases into UT?

SummerSun: UT that tests if the expected exception get thrown? sounds like a good idea

jdneo: Yes

PR #256, opened 2018-07-17
align wording, remove unnecessary check

SummerSun:


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/WarArtifactHandlerImpl.java
```diff
@@ -49,9 +50,7 @@ public void publish(IDeployTargetAdapter deployTarget) throws MojoExecutionExcep
             }
         }
 
-        if (retryCount > DEFAULT_MAX_RETRY_TIMES) {
-            throw new MojoExecutionException(DEPLOY_FAILURE);
-        }
+        throw new MojoExecutionException(DEPLOY_FAILURE);
```
jdneo: String.format()?

PR #257, opened 2018-07-17
update web app samples

jdneo:


docs/web-app-samples.md
```diff
@@ -110,20 +106,11 @@ The following configuration is applicable for below scenario:
...
-                        <targetPath>webapps</targetPath>
-                        <includes>
-                           <include>*.war</include>
-                        </includes>
-                     </resource>
-                  </resources>
+                  <!-- WAR deployment -->
+                  <deploymentType>war</deploymentType>
+
+                  <!-- Specify the war file location, optional if the war file location is: ${project.build.directory}/${project.build.finalName}.war -->
```
SummerSun: How about  "If it is not specified, ${project.build.directory}/${project.build.finalName}.war will be used as default."?

jdneo: Yeah, that's much better

PR #260, opened 2018-07-17
update integration test to cover AUTO deployment type

SummerSun:


jdneo:
If we want to cover the AUTO deployment, I suggest that we remove the `<deploymentType>` in the plugin. This can verify the default behavior of the AUTO deployment

PR #265, opened 2018-07-18
add azure function annotation sample HttpTrigger

JanusWu:
add azure function annotation sample HttpTrigger


jdneo:
Suggest to reference the issue number in the descriptions. For example:

workitem: #...(issue number)

docs/FunctionSamples/HTTP/README.md
```diff
@@ -0,0 +1,9 @@
+### Steps
+1. Open `pom.xml`, replace the value of `resource-group` and `azure-functions-name` with the real value you want.
+2. Run `mvn clean package` under `HTTP` folder to build the project.
+3. Run `mvn azure-functions:run` to run the project locally, if the function app not exists in Azure, it will create a new one.
```
jdneo: not exists -> does not exist

docs/FunctionSamples/HTTP/pom.xml
```diff
@@ -0,0 +1,138 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.microsoft.azure.functions.samples</groupId>
+    <artifactId>azure-functions-samples</artifactId>
```
jdneo: azure-functions-samples-http

docs/FunctionSamples/README.md
```diff
@@ -0,0 +1,19 @@
+# Microsoft Azure Functions in Java
+
+This repo contains Java Azure Functions annotation samples for different events.
+
+## Prerequisites
+- JDK 1.8
+- Maven 3.0+
+- [.NET Core SDK](https://www.microsoft.com/net/learn/get-started/windows)
+- [Azure Functions Core Tools v2](https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local#v2)
+- [Maven Plugin for Azure Functions](https://github.com/Microsoft/azure-maven-plugins/blob/develop/azure-functions-maven-plugin/README.md)
```
jdneo: We should use relative path here, then it will point to the README file in different branches

docs/FunctionSamples/README.md
```diff
@@ -0,0 +1,19 @@
...
+- An active azure subscription
+
+## Azure Functions List and Test Pass
```
jdneo: Any reason to put this section here?

JanusWu: Just want to give a summary of samples under FunctionSamples and test status.

pragnagopa:
Can you add these under https://github.com/Microsoft/azure-maven-plugins/tree/develop/azure-functions-maven-plugin? Create a test folder and add tests there instead of docs folder

pragnagopa:
@jdneo - Can we merge these tests for now? It is ok to run them manually until we have automation in place.

jdneo:
@pragnagopa - We decided to add more integration tests here: https://github.com/Microsoft/azure-maven-plugins/tree/develop/azure-functions-maven-plugin/src/it

Recently we add storage queue trigger & output. Those integration tests will build the project locally, deploy to Azure and trigger the function, finally verify if the result is correct.

The same code can also be used for manual tests. And we will add more test cases to cover all the annotations. I think this could be a better solution in short-term.

PR #292, opened 2018-07-30
update the http integration tests

jdneo:


azure-functions-maven-plugin/src/test/java/com/microsoft/azure/maven/function/invoker/CommonUtils.java
```diff
@@ -84,7 +84,7 @@ public static void runVerification(Runnable verification) throws Exception {
         }
 
         if (i >= RETRY_TIMES) {
-            throw new Exception("Integration test fail for Azure Functions storage-queue-trigger");
+            throw new Exception("Integration test fails for 5 times.");
```
SummerSun: since we have no logic to do after while loop, we could use return in try block to save the if condition check

PR #293, opened 2018-07-30
 implement zipArtifactHandler, add zip deployment type for web app

SummerSun:


azure-maven-plugin-lib/pom.xml
```diff
@@ -147,6 +147,11 @@
             <artifactId>junit</artifactId>
             <version>${junit.version}</version>
         </dependency>
+        <dependency>
+            <groupId>net.lingala.zip4j</groupId>
```
jdneo: Can we use the the original zip api in the JDK instead of adding a dependency here?

See: https://www.mkyong.com/java/how-to-compress-files-in-zip-format/

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
```diff
@@ -32,34 +33,47 @@ public FTPArtifactHandlerImpl(final T mojo) {
...
-        final String outputFolder = this.mojo.getPluginName().replaceAll(MAVEN_PLUGIN_POSTFIX, "");
-        return Paths.get(mojo.getBuildDirectoryAbsolutePath(), outputFolder, this.mojo.getAppName()).toString();
+        return Paths.get(mojo.getBuildDirectoryAbsolutePath(), this.mojo.getAppName()).toString();
     }
 
     @Override
     public void publish(DeployTarget target) throws IOException, MojoExecutionException {
+        if (target.getApp() instanceof WebApp) {
+            prepareResources();
+            assureStageDirectoryNotEmpty();
```
jdneo: Any reason to add this check method here?

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
```diff
@@ -32,34 +33,47 @@ public FTPArtifactHandlerImpl(final T mojo) {
     }
 
     protected String getDeploymentStageDirectory() {
-        final String outputFolder = this.mojo.getPluginName().replaceAll(MAVEN_PLUGIN_POSTFIX, "");
-        return Paths.get(mojo.getBuildDirectoryAbsolutePath(), outputFolder, this.mojo.getAppName()).toString();
+        return Paths.get(mojo.getBuildDirectoryAbsolutePath(), this.mojo.getAppName()).toString();
```
jdneo: Why we remove the plugin specific folder?

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/ZIPArtifactHandler.java
```diff
@@ -6,25 +6,69 @@
...
+    public void publish(DeployTarget target) throws ZipException, MojoExecutionException, IOException {
+        // todo: function app zip deploy could not be tested until the sdk 1.14.0 release
+        prepareResources();
+        assureStageDirectoryNotEmpty();
         target.zipDeploy(getZipFile());
     }
 
-    protected File getZipFile() {
-        return null;
+    protected void prepareResources() throws IOException {
```
jdneo: duplicated codes also found in `FTPArtifactHandlerImpl.java`. Is it because in case the `FTPArtifactHandlerImpl.java` might be removed in future?

azure-webapp-maven-plugin/src/it/14-linux-zip/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+                <artifactId>@project.artifactId@</artifactId>
+                <version>@project.version@</version>
+                <configuration>
+                    <authentication>
+                        <serverId>azure-auth</serverId>
+                    </authentication>
+                    <resourceGroup>maven-webapp-it-rg-14</resourceGroup>
+                    <appName>maven-it-app-${maven.build.timestamp}</appName>
+                    <region>westus</region>
+                    <linuxRuntime>tomcat 8.5-jre8</linuxRuntime>
```
jdneo: The recommended usage should be: 

- tomcat -> warDeploy
- Java SE -> zipDeploy

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
SummerSun: function app does not need copy resources to stage directory

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
SummerSun: as we synced offline, there is no need to use target/azure-webapp/appname as stage directory, just target/appname is ok "azure-webapp" just hard coded in previous code snippet in ftpartifacthandlerimpl

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
jdneo: I mean the method `assureStageDirectoryNotEmpty()`

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
jdneo: `azure-webapp` is needed here:
- It has no harm to have this folder
- Imaging if in the future the plugin needs to generate some files besides the deployment file, where to place them? The folder `azure-webapp` is to specify the scope for our plugin.
- Take the convention of other popular plugins, for example: `findbugs`

azure-maven-plugin-lib/pom.xml
SummerSun: Just compare the zip folder function, in the example #2 in the above link, we will have to write 50+ lines of code to zip a folder. And zip4j makes this happen in 5 lines of code.

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
SummerSun: convention makes good point here but no harm does not means necessary

files besides deployment files should be treated files that not used during deployment, back to our offline discussion, stage directory is within target folder, it is not visible to users, it is just a temp folder created by plugin to stage the files that will deploy to web app

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
SummerSun: jar deploy does not need copy resources neither, it just prepare the jar file to stage directory itself. 

This change use function prepareResources() just do copy resources (if any), and leave resources check to function assureStageDirectoryNotEmpty()

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
SummerSun: Add another reason that I remove this is that, we move the logic to library module. It is totally ok when we hard code 'azure-webapp' or 'function-app' in their own module, but it is kind of not that natural to split out the plugin name to form a temporary folder invisible to users in target folder in library module. We want keep the dependency is plugin module to library module not the reverse.

azure-maven-plugin-lib/pom.xml
jdneo: Adding a new third party dependency is fine. What worth concerning is that:
- Is this dependency a popular one?
- Is this dependency still actively releasing?
- How about the license issue?

For `net.lingala.zip4j`, What I'm concerning is that, I cannot find the source code of it, usage data from the Maven Central is not very high, and it has become stale since 2013. That's why I'm thinking if it is possible to control the logic by our own.

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
jdneo: OK, that's fine. But the error handling may need to be improved.

The string `NO_RESOURCES` contains two scenario:
- No resources were specified in pom.xml
- No resources were copied to stage directory

A better way is to separate these two conditions, and let the user know what's happening.

And for `No resources were specified in pom.xml` condition, do we need to throw an error or just a warning to tell the user the deployment is skipped?

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/ZIPArtifactHandler.java
jdneo: OK

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
```diff
@@ -6,24 +6,26 @@
...
 import java.nio.file.Paths;
 import java.util.List;
 
 public class FTPArtifactHandlerImpl<T extends AbstractAppServiceMojo> implements ArtifactHandler {
     private static final String DEFAULT_WEBAPP_ROOT = "/site/wwwroot";
     private static final int DEFAULT_MAX_RETRY_TIMES = 3;
-    private static final String NO_RESOURCES_CONFIG = "No resources specified in pom.xml. Skip artifacts deployment.";
-    private static final String WEBAPP_PLUGIN_NAME = "azure-webapp-maven-plugin";
     private static final String MAVEN_PLUGIN_POSTFIX = "-maven-plugin";
+    private static final String NO_RESOURCES = "No resources were specified in pom.xml or copied to staging directory.";
```
jdneo: Better for error log here?

SummerSun: https://github.com/Microsoft/azure-maven-plugins/issues/297

Opened a separate issue to discuss and solve it.

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
```diff
@@ -38,28 +40,43 @@ protected String getDeploymentStageDirectory() {
...
+    protected void prepareResources() throws IOException {
+        final List<Resource> resources = this.mojo.getResources();
+
+        if (resources != null && !resources.isEmpty()) {
+            Utils.copyResources(mojo.getProject(), mojo.getSession(),
+                mojo.getMavenResourcesFiltering(), resources, getDeploymentStageDirectory());
+        }
+    }
+
+    protected void assureStageDirectoryNotEmpty() throws MojoExecutionException {
```
jdneo: Please change `stage` to `staging`

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
```diff
@@ -38,28 +40,43 @@ protected String getDeploymentStageDirectory() {
...
+        final String stageDirectory = getDeploymentStageDirectory();
```
jdneo: The variable name is confusing here because of we both have `xxxDirectory` and `xxxFolder`

How about for the Those `String` typed variable, we call `xxxDirectoryPath`

For `File` typed variable, we call `xxxDirectory`

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/ZIPArtifactHandler.java
```diff
@@ -6,25 +6,69 @@
...
+    protected void assureStageDirectoryNotEmpty() throws MojoExecutionException {
+        final String stageDirectory = getDeploymentStageDirectory();
+        final File stageFolder = new File(stageDirectory);
+        final File[] files = stageFolder.listFiles();
+        if (!stageFolder.exists() || !stageFolder.isDirectory() || files == null || files.length == 0) {
+            throw new MojoExecutionException(String.format(NO_RESOURCES, stageDirectory));
+        }
+    }
+
+    protected File getZipFile() throws MojoExecutionException {
```
jdneo: Same problems here: avoid mix use of `Directory` and `Folder`

azure-webapp-maven-plugin/src/it/14-linux-zip/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+                    <artifactId>spring-boot-starter-tomcat</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-jetty</artifactId>
+        </dependency>
+    </dependencies>
+    <build>
```
jdneo: add `<finalName>app</finalName>`

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
```diff
@@ -6,60 +6,77 @@
...
+    private static final String NO_RESOURCES = "No resources were specified in pom.xml or copied to staging directory.";
 
     protected T mojo;
 
     public FTPArtifactHandlerImpl(final T mojo) {
         this.mojo = mojo;
     }
 
-    protected String getDeploymentStageDirectory() {
+    protected String getDeploymentStagingDirectory() {
```
jdneo: `getDeploymentStagingDirectory()` -> `getDeploymentStagingDirectoryPath()`

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
```diff
@@ -6,60 +6,77 @@
...
+        final List<Resource> resources = this.mojo.getResources();
+
+        if (resources != null && !resources.isEmpty()) {
+            Utils.copyResources(mojo.getProject(), mojo.getSession(),
+                mojo.getMavenResourcesFiltering(), resources, getDeploymentStagingDirectory());
+        }
+    }
+
+    protected void assureStagingDirectoryNotEmpty() throws MojoExecutionException {
+        final String stagingDirectory = getDeploymentStagingDirectory();
```
jdneo: `stagingDirectory` -> `stagingDirectoryPath`

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
```diff
@@ -6,60 +6,77 @@
...
+        final File stagingFolder = new File(stagingDirectory);
```
jdneo: `stagingFolder` -> `stagingDirectory`

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/ZIPArtifactHandler.java
```diff
@@ -6,25 +6,69 @@
...
+        final File stagingDirectory = new File(stagingDirectoryPath);
+        final File[] files = stagingDirectory.listFiles();
+        if (!stagingDirectory.exists() || !stagingDirectory.isDirectory() || files == null || files.length == 0) {
+            throw new MojoExecutionException(String.format(NO_RESOURCES, stagingDirectoryPath));
+        }
+    }
+
+    protected File getZipFile() throws MojoExecutionException {
+        final String stagingDirectoryPath = getDeploymentStagingDirectory();
+        final String outputFolder = this.mojo.getPluginName().replaceAll(MAVEN_PLUGIN_POSTFIX, "");
```
jdneo: Can we use `stagingDirectoryPath` here to get the zip file name?

Issue #295, opened 2018-07-31
Null pointer error when <targetPath> not set in <resource> block


SummerSun:
### Expected behavior

Just use site/wwwroot/ to deploy the resource

### Actual behavior

Build failure, null pointer error pops up and users does not get any error message to know what is going wrong.

### Root cause

We just use Path.get to combine the target path ```Paths.get(targetDirectory, resource.getTargetPath()``` without null checking. If resource.getTargetPath() is null, empty string should be used to set resource target path to site/wwwroot/

PR #296, opened 2018-07-31
fix issue: use emtpty string when targetPath not set

SummerSun:


SummerSun:
fix issue #295

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/Utils.java
```diff
@@ -82,7 +82,12 @@ public static void copyResources(final MavenProject project, final MavenSession
                                      final MavenResourcesFiltering filtering, final List<Resource> resources,
                                      final String targetDirectory) throws IOException {
         for (final Resource resource : resources) {
-            resource.setTargetPath(Paths.get(targetDirectory, resource.getTargetPath()).toString());
+            final String targetPath = resource.getTargetPath();
```
jdneo: Can we make it simpler?
```java
final String targetPath = resource.getTargetPath() == null ? "" : resource.getTargetPath();
resource.setTargetPath(Paths.get(targetDirectory, targetPath).toString());
```

SummerSun: sure, but if targetPath is empty string, my change could save a call to Paths.get, that's why I use if block

PR #300, opened 2018-08-01
Add unit tests for FTPArtifactHandlerImpl

SummerSun:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/FTPArtifactHandlerImpl.java
```diff
@@ -38,15 +39,27 @@ protected String getDeploymentStagingDirectoryPath() {
         return Paths.get(mojo.getBuildDirectoryAbsolutePath(), outputFolder, this.mojo.getAppName()).toString();
     }
 
+    protected boolean isPrepareResourceRequired(final DeployTarget target) {
```
jdneo: `isResourcePreparationRequired`

PR #303, opened 2018-08-02
Add zip deployment type in function app

SummerSun:


azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -44,29 +44,27 @@
...
      * Deployment type to deploy Web App. Supported values:
      * <ul>
      * <li>msdeploy</li>
      * <li>ftp</li>
+     * <li>zip</li>
      * </ul>
      *
      * @since 0.1.0
      */
-    @Parameter(property = "functions.deploymentType", defaultValue = MS_DEPLOY)
```
akaroml: So we don't need a default value?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -169,9 +167,11 @@ protected void configureAppSettings(final Consumer<Map> withAppSettings, final M
     //endregion
 
     protected ArtifactHandler getArtifactHandler() {
-        switch (getDeploymentType().toLowerCase(Locale.ENGLISH)) {
```
akaroml: Nice change!

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/configurations/DeploymentType.java
```diff
@@ -0,0 +1,32 @@
...
+public enum DeploymentType {
+    MS_DEPLOY,
+    ZIP,
+    FTP;
+
+    public static DeploymentType fromString(final String input) {
+        if (StringUtils.isEmpty(input)) {
+            return MS_DEPLOY;
+        }
+        switch (input.toUpperCase(Locale.ENGLISH)) {
```
akaroml: An extra blank line between different scopes (if and switch).

azure-functions-maven-plugin/pom.xml
```diff
@@ -195,11 +194,6 @@
             <artifactId>reflections</artifactId>
             <version>${reflections.version}</version>
         </dependency>
-        <dependency>
```
akaroml: Why are we deleting this reference and how is it related to the problem we want to solve in this change list?

SummerSun: We upgraded the azure sdk version to 1.14.0 in library module in the previous PR(#302) which import a  reference to com.fasterxml.jackson.core version 2.9.4 and it conflicts with the version we indirectly referenced through azure-storage.

It causes the functions module build failure and directs us to the problem that we have duplicate dependencies in library module and in functions module. So we remove the dependency in functions module to solve the build failure.

akaroml: This answers why. But it does not relate to the topic of THIS specific change list.
The problem should be addressed in #302 or a follow-up PR dedicated to it.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
akaroml: If the annotation can assign a default value, then using it is preferred than write extra code. An example, when you define a variable, you want to initialize it immediately like this:
int i = 100;
But not:
int i;
...
i = 100;

azure-functions-maven-plugin/pom.xml
SummerSun: Yes, this is not related to this change list. I separate it to PR #304

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
SummerSun: updated

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -17,19 +17,20 @@
...
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
-import org.codehaus.plexus.util.StringUtils;
 
-import java.util.Locale;
 import java.util.Map;
 import java.util.function.Consumer;
 
+
```
jdneo: multiple empty line.

suggest adding

```
<module name="EmptyLineSeparator">
    <property name="allowMultipleEmptyLines" value="false"/>
 </module>
```

in the checkstyle configuration. You can add this in another PR

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -44,29 +45,27 @@
...
      * <ul>
      * <li>msdeploy</li>
      * <li>ftp</li>
+     * <li>zip</li>
      * </ul>
      *
      * @since 0.1.0
      */
-    @Parameter(property = "functions.deploymentType", defaultValue = MS_DEPLOY)
+    @Parameter(property = "functions.deploymentType", defaultValue = "msdeploy")
```
jdneo: use zip as default

akaroml: The common practice is to use enum type internally. Using string is a downgrade.
There should be some auto-converting mechanism (serialization/deserialization) to automatically convert from string to enum type. The result is that we use enum value directly in the function body.

This is the next level of excellence we want to push.

SummerSun: @jdneo  I will make that change in next PR.

@akaroml  I tried, by overriding the toString function of an enum value. But problem here is that, the default value requires a constant. That is the one I didn't figure out. I will keep trying.

PR #307, opened 2018-08-03
Use zip deploy as default in azure function app

SummerSun:


azure-functions-maven-plugin/src/it/0-http-trigger/pom.xml
```diff
@@ -57,7 +57,7 @@
                     <authentication>
                         <serverId>azure-auth</serverId>
                     </authentication>
-                    <resourceGroup>maven-functions-it-rg-0</resourceGroup>
```
SummerSun: update resource group name to align with integration test resource groups

azure-functions-maven-plugin/src/it/0-http-trigger/pom.xml
```diff
@@ -18,7 +18,7 @@
         <functionAppName>maven-functions-it-${timestamp}-0</functionAppName>
         <functionAppRegion>westus</functionAppRegion>
         <stagingDirectory>${project.build.directory}/azure-functions/${functionAppName}</stagingDirectory>
-        <functionResourceGroup>java-functions-group</functionResourceGroup>
```
SummerSun: since we used deploymentType default value in integration test, it will naturally test against zip deploy without any change after we change zip deploy as default value.

azure-functions-maven-plugin/src/it/0-http-trigger/pom.xml
```diff
@@ -57,7 +57,7 @@
                     <authentication>
                         <serverId>azure-auth</serverId>
                     </authentication>
-                    <resourceGroup>maven-functions-it-rg-0</resourceGroup>
+                    <resourceGroup>${functionResourceGroup}</resourceGroup>
```
akaroml: Nice to reference the group name using a variable.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -56,7 +56,7 @@
      *
      * @since 0.1.0
      */
-    @Parameter(property = "functions.deploymentType", defaultValue = DeploymentType.ConstantValues.MS_DEPLOY_VALUE)
+    @Parameter(property = "functions.deploymentType", defaultValue = DeploymentType.ConstantValues.ZIP_VALUE)
```
akaroml: We talked about the v2 schema yesterday. Do we want to use ZIPDEPLOY by default for v1 schema users?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/configurations/DeploymentType.java
```diff
@@ -29,17 +29,17 @@
 
     public static DeploymentType fromString(final String input) {
         if (StringUtils.isEmpty(input)) {
-            return MS_DEPLOY;
+            return ZIP;
```
akaroml: Here to reconsider the fallback logic. When users use this setting entry but with empty or wrong values, we should explicitly show an error but not recover silently. The current logic tolerant wrong input and the wrong configuration will stay without anyone trying to fix because they don't know there's an error.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/configurations/DeploymentType.java
```diff
@@ -29,17 +29,17 @@
...
         switch (input.toLowerCase(Locale.ENGLISH)) {
             case ConstantValues.FTP_VALUE:
                 return FTP;
-            case ConstantValues.ZIP_VALUE:
-                return ZIP;
             case ConstantValues.MS_DEPLOY_VALUE:
-            default:
                 return MS_DEPLOY;
+            case ConstantValues.ZIP_VALUE:
+            default:
```
akaroml: The same concern applies to the default branch. We should throw an error if we cannot process an input.

SummerSun: Agree, just use the default value to proceed the deployment is not a good practice for users to know what is wrong. Issue #308 was created to track it since web app and function app have the same behavior and we need to fix them both.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
SummerSun: I think the default value could align between schema v1 and v2. We will release the zip deploy with the documents telling users about this change, and it will know breaking the deployment experience for current users.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -37,12 +37,14 @@
 public class DeployMojo extends AbstractFunctionMojo {
     public static final String FUNCTION_DEPLOY_START = "Starting deployment to Azure Function App ";
     public static final String FUNCTION_DEPLOY_SUCCESS =
-            "Successfully deployed Azure Function App at https://%s.azurewebsites.net";
+        "Successfully deployed Azure Function App at https://%s.azurewebsites.net";
```
akaroml: "Azure Functions app"

akaroml: The name of the service is "Azure Functions", Please fix all.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -166,15 +168,17 @@ protected void configureAppSettings(final Consumer<Map> withAppSettings, final M
...
+    protected ArtifactHandler getArtifactHandler() throws MojoExecutionException {
         switch (DeploymentType.fromString(getDeploymentType())) {
+            case MS_DEPLOY:
+                return new MSDeployArtifactHandlerImpl(this);
             case FTP:
                 return new FTPArtifactHandlerImpl(this);
             case ZIP:
                 return new ZIPArtifactHandlerImpl(this);
-            case MS_DEPLOY:
+            case UNKNOWN:
```
akaroml: There is no need for an UNKNOWN case because everything else falls into the default branch and is considered an error.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/configurations/DeploymentType.java
```diff
@@ -13,7 +13,8 @@
 public enum DeploymentType {
     MS_DEPLOY(ConstantValues.MS_DEPLOY_VALUE),
     ZIP(ConstantValues.ZIP_VALUE),
-    FTP(ConstantValues.FTP_VALUE);
+    FTP(ConstantValues.FTP_VALUE),
+    UNKNOWN(null);
```
akaroml: There is no need for this.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
SummerSun: I will fix them all in next PR

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/configurations/DeploymentType.java
SummerSun: The above unknown is not necessary, but this one is required for default return value, I have to have some value to switch in DeployMojo.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
jdneo: I prefer using `Azure Function App` here.

Azure Functions is the service name. But for a function instance, the name is `Function App`.

See:

image

akaroml: Good to know that they have an official term `function app`. In this case, we don't need to mention Azure brand every time. So I would suggest `function app` and the message looks like:
Successfully deployed the function app at ...

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -37,9 +37,9 @@
 public class DeployMojo extends AbstractFunctionMojo {
     public static final String FUNCTION_DEPLOY_START = "Starting deployment to Azure Function App ";
     public static final String FUNCTION_DEPLOY_SUCCESS =
-            "Successfully deployed Azure Function App at https://%s.azurewebsites.net";
+        "Successfully deployed Azure Function App at https://%s.azurewebsites.net";
     public static final String FUNCTION_APP_CREATE_START = "Target Azure Function App does not exist. " +
-            "Creating a new Azure Function App ...";
+        "Creating a new Azure Function App ...";
```
akaroml: Creating a new function app...

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -35,14 +35,14 @@
  */
 @Mojo(name = "deploy", defaultPhase = LifecyclePhase.DEPLOY)
 public class DeployMojo extends AbstractFunctionMojo {
-    public static final String FUNCTION_DEPLOY_START = "Starting deployment to Azure Function App ";
+    public static final String FUNCTION_DEPLOY_START = "Starting deployment to function app...";
```
akaroml: If you can decide whether plural form needs to be used, that's good. If not, use the following.
Deploying function app(s)...

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -35,14 +35,14 @@
...
     public static final String FUNCTION_DEPLOY_SUCCESS =
-            "Successfully deployed Azure Function App at https://%s.azurewebsites.net";
-    public static final String FUNCTION_APP_CREATE_START = "Target Azure Function App does not exist. " +
-            "Creating a new Azure Function App ...";
-    public static final String FUNCTION_APP_CREATED = "Successfully created Azure Function App ";
-    public static final String FUNCTION_APP_UPDATE = "Updating Azure Function App...";
-    public static final String FUNCTION_APP_UPDATE_DONE = "Successfully updated Azure Function App ";
+        "Successfully deployed function app at https://%s.azurewebsites.net.";
```
akaroml: Successfully deployed function app(s) at...

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -35,14 +35,14 @@
...
+    public static final String FUNCTION_APP_CREATE_START = "Target function app does not exist. " +
```
akaroml: The specified function app does not exist.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -35,14 +35,14 @@
...
+        "Creating a new function app...";
+    public static final String FUNCTION_APP_CREATED = "Successfully created function app.";
```
akaroml: a function app or function app(s)

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -35,14 +35,14 @@
...
+    public static final String FUNCTION_APP_UPDATE = "Updating function app...";
```
akaroml: Updating the specified function app...

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -35,14 +35,14 @@
...
+    public static final String FUNCTION_APP_UPDATE_DONE = "Successfully updated function app.";
```
akaroml: the function app.

PR #314, opened 2018-08-06
extract common logic into ArtifactHandlerBase

jdneo:


SummerSun:
Remove ``getDeploymentStageDirectory``` from AbstractFunctionMojo, let it call the one defined in parent class and remove the related unit tests to library module.

jdneo:
Tracking: https://github.com/Microsoft/azure-maven-plugins/issues/315

PR #317, opened 2018-08-06
Move getStagingDirectoryPath() from FunctionMojo to AppServiceMojo

### LINKED ISSUE #315: Move getDeploymentStageDirectory() from AbstractFunctionMojo to AbstractAppServiceMojo
This is a very common logic method and should be shared between different mojos

jdneo:
resolve #315


azure-maven-plugin-lib/src/test/java/com/microsoft/azure/maven/AbstractAppServiceMojoTest.java
```diff
@@ -0,0 +1,46 @@
...
+    private AbstractAppServiceMojo mojo = new AbstractAppServiceMojo() {
+        @Override
+        protected void doExecute() throws Exception {
+
+        }
+    };
+
+    @Test
+    public void getDeploymentStagingDirectory() {
+        final AbstractAppServiceMojo spy = spy(mojo);
```
SummerSun: better naming mojoSpy

PR #321, opened 2018-08-06
Throw exception when user set unknown deployment type

SummerSun:


azure-webapp-maven-plugin/src/test/resources/pom-linux.xml
```diff
@@ -25,6 +25,7 @@
                             <value>80</value>
                         </property>
                     </appSettings>
+                    <deploymentType>auto</deploymentType>
```
jdneo: Why we need explicitly set deployment type to auto here?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/AbstractWebAppMojo.java
```diff
@@ -329,7 +330,12 @@ public void setDeploymentSlot(DeploymentSlotSetting slotSetting) {
         map.put(JAVA_WEB_CONTAINER_KEY, getJavaWebContainer().toString());
         map.put(LINUX_RUNTIME_KEY, StringUtils.isEmpty(linuxRuntime) ? "" : linuxRuntime);
         map.put(DOCKER_IMAGE_TYPE_KEY, WebAppUtils.getDockerImageType(getContainerSettings()).toString());
-        map.put(DEPLOYMENT_TYPE_KEY, getDeploymentType().toString());
+
+        try {
+            map.put(DEPLOYMENT_TYPE_KEY, getDeploymentType().toString());
+        } catch (MojoExecutionException e) {
+            getLog().info(DeploymentType.UNKNOWN_DEPLOYMENT_TYPE);
```
jdneo: print this into debug() channel since this is telemetry related code.

azure-webapp-maven-plugin/src/test/resources/pom-linux.xml
SummerSun: testing purpose

PR #357, opened 2018-08-21
Throw exception early if <resource> is required but not configured

SummerSun:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/artifacthandler/ArtifactHandlerBase.java
```diff
@@ -32,12 +32,13 @@ protected void assureStagingDirectoryNotEmpty() throws MojoExecutionException {
...
 
-    protected void prepareResources() throws IOException {
+    protected void prepareResources() throws IOException, MojoExecutionException {
         final List<Resource> resources = this.mojo.getResources();
-
-        if (resources != null && !resources.isEmpty()) {
-            Utils.copyResources(mojo.getProject(), mojo.getSession(),
-                    mojo.getMavenResourcesFiltering(), resources, mojo.getDeploymentStagingDirectoryPath());
+        if (resources == null || resources.isEmpty()) {
+            throw new MojoExecutionException("Please configure <resources> in pom.xml.");
```
jdneo: The configuration of <resources> is empty, please make sure it isconfigured in pom.xml.

PR #359, opened 2018-08-21
Add schemaVersion to prepare the re-construction of setting

SummerSun:


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/HandlerFactoryImpl.java
```diff
@@ -68,13 +81,31 @@ public RuntimeHandler getRuntimeHandler(final AbstractWebAppMojo mojo) throws Mo
         throw new MojoExecutionException(NO_RUNTIME_HANDLER);
     }
 
+    protected RuntimeHandler getRuntimeHandlerFromV2Schema(final AbstractWebAppMojo mojo)
+        throws MojoExecutionException {
+        // todo
+        throw new MojoExecutionException("Unimplemented for schema version: " + mojo.getSchemaVersion());
```
jdneo: Unavailable schema version: ....

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/HandlerFactoryImpl.java
```diff
@@ -94,6 +125,12 @@ public ArtifactHandler getArtifactHandler(final AbstractWebAppMojo mojo) throws
         }
     }
 
+    protected ArtifactHandler getArtifactHandlerFromV2Schema (final AbstractWebAppMojo mojo)
+        throws MojoExecutionException {
+        // todo
+        throw new MojoExecutionException("Unimplemented for schema version: " + mojo.getSchemaVersion());
```
jdneo: Unavailable schema version: ....

SummerSun: Will replace the todo in next PR of implementation, do you think we still need to update for the words?

jdneo: @SummerSun Please also involve @Matthew-Dong to discuss about this change. Since we may also want to release the new plugin to support zip deploy ASAP. So it is worth discussing whether we need to include this change into the release.

SummerSun: I think we all agree that this will not be included in the zip deploy.

Issue #354, opened 2018-08-20
Refactoring <runtime> settings


SummerSun:
New settings of runtime after discussion in #200 will be set as following:

<runtime>element | Description | Required? | Default value
-- | -- | -- | --
os | Specifies the OS of the Web App, options are: windows/linux/docker | Yes | No default
javaVersion | Specifies the JVM version for your Web App | Yes | No default
webContainer | Specifies the web container of your Web App | No | tomcat 8.5
image | Specifies the Docker image name | No | No default
serverId | Specifies the credentials for private docker hub images or private container registry images | No | No default
registryUrl | Specifies the URL of private container registry images | No | No default

SummerSun:
Degisn discussion:

New configuration ```<schemaVersion>``` indicate which version of the schema will be used to deploy the Web App. We need different handlers for different schema versions. And we will keep using the Handler Factory to create them.

The current design of handlers have strong coupling relationship with mojo. It is heavy to pass around mojo from functions to functions and it makes it hard for the handlers to be reused for other kind of framework. From V1 handlers, we will add a middle level of Model to decouple them.

Considering that there are a lot of properties required from mojo, we will use Builder Pattern to handle it. Then if other kind of framework would like to re-use runtime handlers in the future, it could just fill in all the properties of the BaseRuntimeHandler from its own "mojo" and the Builder could generate the same handler in the same way.

We will create an abstract base class **BaseRuntimeHandler** with a generic type of Builder. 

```java
public abstract class BaseRuntimeHandler implements RuntimeHandler {
    // required properties of runtime handlers
    public abstract static class Builder<T extends Builder<T>> {
        public abstract BaseRuntimeHandler build();

        protected abstract T self();
    }
}
```

Then HandlerFactory will use the mojo to create different runtime handlers of different scenarios.

```java
switch (OperatingSystemEnum.fromString(runtime.getOs())) {
            case Windows:
                return new WindowsRuntimeHandlerImplV2.Builder()
                    .appName(mojo.getAppName())
                    .resourceGroup(mojo.getResourceGroup())
                    .region(mojo.getRegion())
                    .pricingTier(mojo.getPricingTier())
                    .servicePlanName(mojo.getAppServicePlanName())
                    .servicePlanResourceGroup((mojo.getAppServicePlanResourceGroup()))
                    .azure(mojo.getAzureClient())
                    .log(mojo.getLog())
                    .build();
            case Linux:
[...]

PR #376, opened 2018-08-31
Implement V2 runtime handlers

SummerSun:
- Determine the runtime handlers by schemaVersion

- Decouple the V2 runtime handlers from mojo

- User Builder Pattern to handler the plenty of parameters


jdneo:
I like the builder pattern here, but one problem I observed is that we have a lot of duplicate code in this PR.

Is it possible that we have only one builder? And this builder has the ability to build different kind of runtime handler from its `build()` method.

SummerSun:
@jdneo Yes, there are lots of duplicate code. Update to use a variable to hold the Builder:
```
        final BaseRuntimeHandler.Builder builder;

        switch (OperatingSystemEnum.fromString(runtime.getOs())) {
            case Windows:
                builder = new WindowsRuntimeHandlerImplV2.Builder();
                break;
            case Linux:
                builder = new LinuxRuntimeHandlerImplV2.Builder();
                break;
            case Docker:
                builder = getV2DockerRuntimeHandlerBuilder(mojo);
                break;
            ...
        }
        return builder.runtime.appName....build();
```

SummerSun:
#354

PR #410, opened 2018-10-02
Mockito upgrade

mureinik:
Upgrade the Mockito dependency to the latest available version on maven-central, 2.22.0 in order to consume its latest bugfixes and performance improvements and in order to make a potential upgrade to JUnit Jupiter easier.

By default, Mockito 2.22.0 is stricter than the previously used 2.4.0. Besides upgrading, this PR contains a minimal set of clean-ups (e.g., removing redundant spying) required to run the test-suite without error. While these changes suggest some future similar changes can also be performed, they are out of scope for this PR.


jdneo:
Hi @mureinik,

We are releasing the new version of plugins these days. The PR will be merged once the new plugins are all released.

Thank you for your contribution.

PR #431, opened 2018-10-22
Update the default value of region to West Europe in V1 schema

SummerSun:


jdneo:
Also update the docs?

PR #432, opened 2018-10-22
 Make region a required configuration for web app in schema version v2

SummerSun:


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/HandlerFactoryImpl.java
```diff
@@ -122,6 +119,12 @@ protected RuntimeHandler getV2RuntimeHandler(final AbstractWebAppMojo mojo)
             .build();
     }
 
+    private void assureV2RequiredPropertyConfigured(final AbstractWebAppMojo mojo) throws MojoExecutionException {
+        if (StringUtils.isEmpty(mojo.getRegion())) {
+            throw new MojoExecutionException("No <runtime> is specified, please configure it in pom.xml.");
```
jdneo: Region or runtime?

PR #434, opened 2018-10-22
Move Parent element after artifact

pragnagopa:
Rearranging parent and artifact to help with version string parsing

@Flanker32 /@jdneo - I ran into this when trying to add CI for azure-functions-java-library and following pattern you used to get the version string from pom [here](https://github.com/Microsoft/azure-maven-plugins/blob/develop/appveyor_function_e2e.yml#L34) 

This change will help with keeping ci build scripts consistent


jdneo:
There must be another way to achieve this. Put parent fields after artifact fields does not like the right convention in the Maven world.

pragnagopa:
Do you have any other suggestions?

jdneo:
Something like:
`<artifactId>azure-functions-java-library<\/artifactId>\s<version>(.*)<\/version>`

Issue #355, opened 2018-08-20
Refactoring <deployment> settings


SummerSun:
New settings of ```<deployment>``` after discussion in #200 will be set as following:

<deployment>element | Description | Required? | Default value
-- | -- | -- | --
resources | Specifies the artifacts to be deployed to your Web App | No | No default

<resources>element | Description | Required? | Default value
-- | -- | -- | --
directory | Specifies where the resources are stored | No | No default
targetPath | Specifies the target path where the resources will be deployed to.This is a relative path to the /site/wwwroot/ folder of FTP server in your Web App | No | No default
includes | A list of patterns to include, e.g. **/*.war | No | No default
excludes | A list of patterns to exclude, e.g. **/*.xml | No | No default

PR #438, opened 2018-10-24
Add unit test for artifact handler v2

SummerSun:
Testing part of work for #355 
- Better naming ArtifactHandlerV2 to ArtifactHandlerImplV2 to avoid confusion
- Add unit tests


azure-webapp-maven-plugin/src/test/resources/artifacthandlerv2/app.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
jdneo: rename the file name to `test.html` to indicate that it's used for test purpose

Issue #437, opened 2018-10-23
Make runtime configuration optional if users specify an existing web app


SummerSun:
The operation "updating web app" is time-consuming and maybe not necessary. We should provide the option for users to just do the deployment and do not update the web app.

In V2 schema version, if users specify an existing web app, the `runtime` configuration is optional, and we skip the updating web app step to run the `deploy` faster.

PR #440, opened 2018-10-24
Skip updating web app if runtime not configured in v2 schema version

SummerSun:
#437

### DIFF
--- a/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/WebAppUtils.java
+++ b/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/WebAppUtils.java
@@ -18,6 +18,8 @@ import com.microsoft.azure.management.appservice.WebApp.DefinitionStages.WithCre
 import com.microsoft.azure.management.appservice.WebApp.DefinitionStages.WithDockerContainerImage;
 import com.microsoft.azure.maven.utils.AppServiceUtils;
 import com.microsoft.azure.maven.webapp.configuration.DockerImageType;
+import com.microsoft.azure.maven.webapp.configuration.RuntimeSetting;
+import com.microsoft.azure.maven.webapp.configuration.SchemaVersion;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.logging.Log;
 import org.codehaus.plexus.util.StringUtils;
@@ -71,7 +73,7 @@ public class WebAppUtils {
     private static void assureLinuxPlan(final AppServicePlan plan) throws MojoExecutionException {
         if (!plan.operatingSystem().equals(OperatingSystem.LINUX)) {
             throw new MojoExecutionException(String.format(SERVICE_PLAN_NOT_APPLICABLE,
-                    plan.name(), OperatingSystem.LINUX.name()));
+                plan.name(), OperatingSystem.LINUX.name()));
         }
     }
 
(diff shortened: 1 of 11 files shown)


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/WebAppUtils.java
```diff
@@ -161,6 +163,20 @@ public static RuntimeStack getLinuxRunTimeStack(String imageName) throws MojoExe
...
+            return false;
+        }
+
+        return !StringUtils.isEmpty(runtime.getOs()) || runtime.getJavaVersion() != null ||
+            runtime.getWebContainer() != null || !StringUtils.isEmpty(runtime.getImage()) ||
+            !StringUtils.isEmpty(runtime.getServerId()) || !StringUtils.isEmpty(runtime.getRegistryUrl());
+    }
+
     /**
      * Work Around:
```
jdneo: `Workaround`, not `Work Around`

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/WebAppUtils.java
```diff
@@ -161,6 +163,20 @@ public static RuntimeStack getLinuxRunTimeStack(String imageName) throws MojoExe
...
+    public static boolean isUpdateWebAppNecessary(final String schemaVersion, final RuntimeSetting runtime)
+        throws MojoExecutionException {
+        if (SchemaVersion.V2 != SchemaVersion.fromString(schemaVersion)) {
+            return true;
+        }
+        if (runtime == null) {
+            return false;
+        }
+
+        return !StringUtils.isEmpty(runtime.getOs()) || runtime.getJavaVersion() != null ||
```
jdneo: `StringUtils.isNotEmpty()`

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/WebAppUtils.java
```diff
@@ -161,6 +163,20 @@ public static RuntimeStack getLinuxRunTimeStack(String imageName) throws MojoExe
         throw new MojoExecutionException(IMAGE_NOT_GIVEN);
     }
 
+    public static boolean isUpdateWebAppNecessary(final String schemaVersion, final RuntimeSetting runtime)
```
jdneo: What if the runtime setting is just the same as the actual runtime of the Web App

SummerSun: Will still do the update. If there are any improvement space, I think it should be handled in the API side.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/WebAppUtils.java
```diff
@@ -161,8 +163,22 @@ public static RuntimeStack getLinuxRunTimeStack(String imageName) throws MojoExe
...
+        throws MojoExecutionException {
+        if (SchemaVersion.V2 != SchemaVersion.fromString(schemaVersion)) {
+            return true;
+        }
+        if (runtime == null) {
+            return false;
+        }
+
+        return StringUtils.isNotEmpty(runtime.getOs()) || runtime.getJavaVersion() != null ||
```
jdneo: Add a util method: `isEmpty()` in runtime

PR #441, opened 2018-10-25
Java version should not be required since we have docker kind of os

SummerSun:
Updated in the description too https://github.com/Microsoft/azure-maven-plugins/issues/354


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/HandlerFactoryImpl.java
```diff
@@ -94,10 +94,6 @@ protected RuntimeHandler getV2RuntimeHandler(final AbstractWebAppMojo mojo)
 
         final BaseRuntimeHandler.Builder builder;
         final RuntimeSetting runtime = mojo.getRuntime();
-        if (StringUtils.isEmpty(runtime.getOs()) || null == runtime.getJavaVersion()) {
```
SummerSun: The `OperatingSystemEnum` will handle it when the os is null or empty.

SummerSun:
Drop this to do a refactoring of all the configuration.

Issue #417, opened 2018-10-11
Show error if building extensions.csproj fails


pragnagopa:
Build should fail if building extensions.csproj fails or cannot build due to missing azure functions core tools
As of now here is the message
```
WARNING] Failed to get Azure Functions Core Tools version locally
[WARNING] Local Azure Functions Core Tools does not support extension auto-install, skip it in the package phase.
[WARNING] Local version of Azure Functions Core Tools (null) does not match the latest (2.0.3). Please update it for the best experience. See: https://aka.ms/azfunc-install
[INFO] Function extension installation done.
[INFO] Successfully built Azure Functions.
[INFO]
[INFO] --- maven-resources-plugin:3.1.0:copy-resources (copy-resources) @ azure-functions-java-endtoendtests ---
```

which is misleading.

pragnagopa:
cc: @jdneo @Flanker32

Flanker32:
@pragnagopa Project could package and deploy to azure without azure function core tools, so could we show a warning and tell user Function could not run locally without azure function core tools instead of showing build fail?

Flanker32:
@pragnagopa  Any comments on this issue?

pragnagopa:
Core Tools is part of requirements for using maven plugin. For functions that do need extensions, for example a queue trigger, funciton.json is not generated and there is no indication that build failed. If users go ahead and deploy without locally testing - again core tools is needed for this as well, it is too late to notify
Being more explicit that core tools is required is better experience

pragnagopa:
Sounds good! Only throw error if extensions are needed

jdneo:
Great, @pragnagopa could you please provide the list of functions that do not need extensions?

pragnagopa:
Except httptrigger, all other triggers need extensions

PR #443, opened 2018-10-25
Show error message when package without function core tools

Flanker32:
Show error message when package without function core tools, for issue #417


azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/handlers/FunctionCoreToolsHandlerImpl.java
```diff
@@ -8,23 +8,31 @@
...
     public static final String INSTALL_FUNCTION_EXTENSIONS_FAIL = "Failed to install the Function extensions";
+    public static final String NO_LOCAL_FUNCTION_CORE_TOOLS = "Local Azure Functions Core Tools does not " +
+            "exist, please install Azure Function Core Tools and try again. See: https://aka.ms/azfunc-install";
     public static final String OUTDATED_LOCAL_FUNCTION_CORE_TOOLS = "Local Azure Functions Core Tools does not " +
-            "support extension auto-install, skip it in the package phase.";
+            "exist or it is too old to support extension auto-install, skip it in the package phase. To install or" +
+            " upgrade it, see: https://aka.ms/azfunc-install";
     public static final String NEED_UPDATE_FUNCTION_CORE_TOOLS = "Local version of Azure Functions Core Tools (%s) " +
-            "does not match the latest (%s). Please update it for the best experience. " + 
+            "does not match the latest (%s). Please update it for the best experience. " +
```
jdneo: Is this String `NEED_UPDATE_FUNCTION_CORE_TOOLS` used in any place?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/handlers/FunctionCoreToolsHandlerImpl.java
```diff
@@ -34,22 +42,27 @@ public FunctionCoreToolsHandlerImpl(final AbstractFunctionMojo mojo, final Comma
...
 
-            if (isLocalVersionSupportAutoInstall(localVersion)) {
-                installFunctionExtension();
+        if (isLocalVersionSupportAutoInstall(localVersion)) {
+            checkVersion(localVersion);
+            installFunctionExtension();
+        } else {
+            if (checkExtensionNecessity(bindingTypes)) {
+                this.mojo.error(NO_LOCAL_FUNCTION_CORE_TOOLS);
+                throw new Exception(NO_LOCAL_FUNCTION_CORE_TOOLS);
```
jdneo: What's the output if we going to this if-block?

Will the `NO_LOCAL_FUNCTION_CORE_TOOLS` be printed twice?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
```diff
@@ -273,12 +274,19 @@ protected FunctionCoreToolsHandler getFunctionCoreToolsHandler(final CommandHand
         return new FunctionCoreToolsHandlerImpl(this, commandHandler);
     }
 
-    protected void installExtension(final FunctionCoreToolsHandler handler) throws Exception {
+    protected void installExtension(final FunctionCoreToolsHandler handler,
+                                    Map<String, FunctionConfiguration> configMap) throws Exception {
         info("");
         info(INSTALL_EXTENSIONS);
-        handler.installExtension();
+        handler.installExtension(getFunctionBindingClasses(configMap));
```
jdneo: Instead of passing the `configMap` deeply. I prefer to have a util method to check if installing the extension is needed before trigger the installing command

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/handlers/FunctionCoreToolsHandlerImpl.java
```diff
@@ -60,38 +53,47 @@ protected void installFunctionExtension() throws Exception {
         );
     }
 
-    protected boolean isLocalVersionSupportAutoInstall(final String localVersion) {
+    protected boolean checkLocalCoreToolsVersion() {
```
jdneo: Let's change the function to `void assureRequirementAddressed()`

And do three steps in it:
- check function cli installed
- check function cli has the ability to install extensions
- check function cli is not out-dated

Issue #445, opened 2018-10-29
Update distributionManagement in function pom to publish to SNAPSHOT repo


Flanker32:
Add distributionManagement infos to function pom to publish to SNAPSHOT repo
```
  <distributionManagement>
    <snapshotRepository>
      <id>ossrh</id>
      <name>Sonatype Snapshots</name>
      <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
      <uniqueVersion>true</uniqueVersion>
      <layout>default</layout>
    </snapshotRepository>
  </distributionManagement>
```

Flanker32:
Solved in #446

PR #446, opened 2018-10-29
Update pom for publish snapshot

Flanker32:
Add <distributionManagement> to pom and remove findbugs-maven-plugin for publish snapshot.
For issue  #445


pom.xml
```diff
@@ -25,6 +25,16 @@
         <tag>HEAD</tag>
     </scm>
 
+    <distributionManagement>
```
jdneo: Do we really need to add `<distributionManagement>` in the parent pom?

Flanker32: For now we need to add this attribute, or jenkins job will throw exception when we release snapshot versions.
``` powershell
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:2.7:deploy (default-deploy) on project azure-maven-plugins: Deployment failed: repository element was not specified in the POM inside distributionManagement element or in -DaltDeploymentRepository=id::layout::url parameter -> [Help 1]
```
We may combine the Aggregate POM and Parent POM , tracking in issue #453

PR #447, opened 2018-10-29
Rename it resource group in case of java lib issue

Flanker32:
Rename it resource group in case of java lib issue


SummerSun:
This is a workaround for https://github.com/Azure/azure-libraries-for-java/issues/536 in our integration test.

Issue #420, opened 2018-10-13
[Functions]Consume CosmosDB annotation updates


pragnagopa:
- Updated CosmosDBInput, CosmosDBOutput and CosmosDBTigger annotations in PR#https://github.com/Azure/azure-functions-java-library/pull/54

- Deleted unsupported annotation NotificationHubOutput

- Published v1.0.0-beta-6 to sonatype
https://oss.sonatype.org/service/local/staging/repository/commicrosoftazure-2655 

cc: @jdneo , @Flanker32

pragnagopa:
Note this also includes EventHubTrigger annotation that supports cardinality

Flanker32:
@pragnagopa  So should we do something for the first issue, or we just need to remove the support of NotificationHubOutput?

pragnagopa:
Yes. Please delete.

Flanker32:
@pragnagopa  We remove the support for NotificationHubOutput in this pr https://github.com/Microsoft/azure-maven-plugins/pull/424, but we found NotificationHubOutput is still exist in [Java library]( https://github.com/Azure/azure-functions-java-library/blob/dev/src/main/java/com/microsoft/azure/functions/annotation/NotificationHubOutput.java ), is there anything wrong with it?

pragnagopa:
Published latest to https://oss.sonatype.org/service/local/staging/repository/commicrosoftazure-2662 
Please ensure to add any annotations with default values to function.json

Flanker32:
@pragnagopa  So for CosmosDB(Trigger,Input,Output), what kind of attributes should we add to function.json, can you share us an example of function.json?

pragnagopa:
Here is function.json form CSharp templates https://github.com/Azure/azure-functions-templates/blob/dev/Functions.Templates/Templates/CosmosDBTrigger-CSharp/function.json
Looks like plugin already generates in the right format. Only action item here is to consume latest library. I will update this issue once snapshot version is published

pragnagopa:
Ping!

Flanker32:
@pragnagopa  Support for NotificationHubOutput has been removed in https://github.com/Microsoft/azure-maven-plugins/pull/424, update pom to use java function lib 1.0.0-beta-7-SNAPSHOT is in https://github.com/Microsoft/azure-maven-plugins/pull/423

Is there any thing else with this issue?

pragnagopa:
Please verify function.json is generated properly when new properties are specified

[...]

PR #449, opened 2018-10-29
Support new attributes in CosmosDB annotations

Flanker32:
Support new attributes in CosmosDB annotations, for issue #420

### DIFF
--- a/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/CosmosDBBinding.java
+++ b/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/CosmosDBBinding.java
@@ -15,7 +15,9 @@ import com.microsoft.azure.functions.annotation.CosmosDBTrigger;
 @JsonInclude(JsonInclude.Include.NON_DEFAULT)
 public class CosmosDBBinding extends BaseBinding {
     public static final String COSMOS_DB_TRIGGER = "cosmosDBTrigger";
-    public static final String COSMOS_DB = "cosmosdb";
+    public static final String COSMOS_DB = "cosmosDB";
+
+    private int collectionThroughput = -1;
 
     private String databaseName = "";
 
@@ -29,18 +31,21 @@ public class CosmosDBBinding extends BaseBinding {
 
     private String connectionStringSetting = "";
 
-    private boolean createIfNotExists = false;
+    private String partitionKey = "";
 
-    private boolean createLeaseCollectionIfNotExists = false;
+    private String createIfNotExists = "false";
+
+    private String createLeaseCollectionIfNotExists = "true";
 
     public CosmosDBBinding(final CosmosDBInput dbInput) {
         super(dbInput.name(), COSMOS_DB, Direction.IN, dbInput.dataType());
 
         databaseName = dbInput.databaseName();
         collectionName = dbInput.collectionName();
-        id = dbInput.id();
-        sqlQuery = dbInput.sqlQuery();
         connectionStringSetting = dbInput.connectionStringSetting();
+        id = dbInput.id();
+        partitionKey = dbInput.partitionKey();
+        sqlQuery = dbInput.sqlQuery();
     }
 
     public CosmosDBBinding(final CosmosDBOutput dbOutput) {
@@ -48,18 +53,20 @@ public class CosmosDBBinding extends BaseBinding {
 
         databaseName = dbOutput.databaseName();
         collectionName = dbOutput.collectionName();
+        createIfNotExists = dbOutput.createIfNotExists() ? "true" : "false";
         connectionStringSetting = dbOutput.connectionStringSetting();
-        createIfNotExists = dbOutput.createIfNotExists();
+        partitionKey = dbOutput.partitionKey();
+        collectionThroughput = dbOutput.collectionThroughput();
     }
 
     public CosmosDBBinding(final CosmosDBTrigger dbTrigger) {
         super(dbTrigger.name(), COSMOS_DB_TRIGGER, Direction.IN, dbTrigger.dataType());
 
+        connectionStringSetting = dbTrigger.connectionStringSetting();
         databaseName = dbTrigger.databaseName();
         collectionName = dbTrigger.collectionName();
         leaseCollectionName = dbTrigger.leaseCollectionName();
-        connectionStringSetting = dbTrigger.connectionStringSetting();
-        createLeaseCollectionIfNotExists = dbTrigger.createLeaseCollectionIfNotExists();
+        createLeaseCollectionIfNotExists = dbTrigger.createLeaseCollectionIfNotExists() ? "true" : "false";
     }
 
     @JsonGetter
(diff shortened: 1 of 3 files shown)


Flanker32:
@pragnagopa 
The generated function.json has been verified according to https://github.com/Azure/azure-functions-templates/blob/1e4c65a5a9db8e1b2f2d87a7c7d03463f4e626f3/Functions.Templates/Bindings/bindings.json#L2122. Now the function plugin could package and deploy cosmos db function, but i tried to write something with CosmosDBOutput, it shows 
``` powershell
2018-10-29T09:07:38.129 [Error] Executed 'Functions.HttpTrigger-Java' (Failed, Id=ebbfaf1b-4816-4314-b62b-6c98c3f139a8)

Message: {"Errors":["One of the specified inputs is invalid"]}

ActivityId: c70f9511-031c-4ac4-abea-095a7b85f849, Request URI: /apps/c2ab109d-a398-4eec-b7aa-7b6c4ce540dd/services/d04992c4-96ae-40bf-82e4-e969510ecf5d/partitions/38719f72-36de-4870-b1ab-3b9a463cf859/replicas/131835576035187869p/, RequestStats: 

RequestStartTime: 2018-10-29T09:07:38.0522210Z, Number of regions attempted: 1

, SDK: Microsoft.Azure.Documents.Common/2.0.0.0, Windows/10.0.14393 documentdb-netcore-sdk/2.0.0
```

And here is the trigger 
``` java
    @FunctionName("HttpTrigger-Java")
    @CosmosDBOutput(name = "output", databaseName = "ToDoList", collectionName = "Items", connectionStringSetting = "AzureCosmosDBConnection", dataType="string")
    public String run(
            @HttpTrigger(name = "req", methods = {HttpMethod.GET, HttpMethod.POST}, authLevel = AuthorizationLevel.ANONYMOUS) HttpRequestMessage<Optional<String>> request,
            final ExecutionContext context) {
        context.getLogger().info("Java HTTP trigger processed a request.");

        // Parse query parameter
        String query = request.getQueryParameters().get("name");
        String name = request.getBody().orElse(query);

        String result = "{ id: " + System.currentTimeMillis() + ", Description: " + name + " }";
        context.getLogger().info(result);
        return result;

    }
```

Do you have any suggestion about this issue?

pragnagopa:
You are receiving a valid error. CosmosDB expects an Id to be set to a GUID.
Following works
```
String result = "{\"id\":\""+UUID.randomUUID()+ "\",\"Description\":\""+name+"\"}";
```
For cosmosDb a better example would be to use POJO as specifying 
```
dataType="string"
```
is not very useful as inserting into documentdb needs an id
Please see example here for reference: https://github.com/Azure/azure-functions-java-worker/blob/dev/endtoendtests/src/main/java/com/microsoft/azure/functions/endtoend/CosmosDBTriggerTests.java#L115

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/CosmosDBBinding.java
```diff
@@ -88,17 +95,27 @@ public String getConnectionStringSetting() {
     }
 
     @JsonGetter
-    public boolean isCreateIfNotExists() {
+    public String isCreateIfNotExists() {
```
pragnagopa: Any reason why using String instead of Boolean?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/CosmosDBBinding.java
```diff
@@ -88,17 +95,27 @@ public String getConnectionStringSetting() {
...
         return createIfNotExists;
     }
 
     @JsonGetter
-    public boolean isCreateLeaseCollectionIfNotExists() {
+    public String isCreateLeaseCollectionIfNotExists() {
```
pragnagopa: Same here any reason why using String instead of Boolean?

azure-functions-maven-plugin/src/test/java/com/microsoft/azure/maven/function/handlers/AnnotationHandlerImplTest.java
```diff
@@ -93,7 +93,8 @@ public void cosmosDBTriggerMethod(@CosmosDBTrigger(name = "cosmos",
                                                 databaseName = "db",
                                                 collectionName = "cl",
                                                 connectionStringSetting = "conn",
-                                                leaseCollectionName = "lease") String in) {
+                                                leaseCollectionName = "lease",
+                                                createLeaseCollectionIfNotExists = false) String in) {
```
pragnagopa: Can you please add test that sets all the properties on cosmos db trigger and verfies generated function.json?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/CosmosDBBinding.java
Flanker32: If bool values false, it will not appear in the generated function.json,  and some required attributes(E.g createIfNotExists) will miss and function will not work in azure.

Flanker32:
@pragnagopa  For now we add attributes according to  [binding.json](https://github.com/Azure/azure-functions-templates/blob/dev/Functions.Templates/Bindings/bindings.json#L2122) here, add there are lots of new attributes add on your pr#https://github.com/Azure/azure-functions-java-library/pull/54, so should we add all new attributes to function.json, is there any new example here? For the [C# example]( https://github.com/Azure/azure-functions-templates/blob/dev/Functions.Templates/Templates/CosmosDBTrigger-CSharp/function.json) hasn't been updated for a long time.

pragnagopa:
Unit test that verifies function.json is created with all the values 
Specified in the annotation is enough. I don’t have a sample that uses all the attributes. Will look for it.

jdneo:
Hi @pragnagopa,

Currently the CosmosDB annotation in `azure-functions-java-library` has more attributes than we have in the plugin: https://github.com/Microsoft/azure-maven-plugins/blob/develop/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/CosmosDBBinding.java

Should we add these newly added attributes as @Flanker32 listed in the `CosmosDBBinding.java` first?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/CosmosDBBinding.java
```diff
@@ -12,54 +12,81 @@
...
-    private String sqlQuery = "";
-
-    private String connectionStringSetting = "";
-
-    private boolean createIfNotExists = false;
-
-    private boolean createLeaseCollectionIfNotExists = false;
+    public static final String COSMOS_DB = "cosmosDB";
+
+    private Integer collectionThroughput = null;
```
jdneo: Add comments to classify which are common attributes, which are trigger attributes, etc...

PR #464, opened 2018-11-02
Set java version to 1.8 when deploy

Flanker32:
Set java version to 1.8 when deploy


jdneo:
@pragnagopa 

Please help take a look when you have time. Looks like the web container also needs to be set since it's a windows app service.

pragnagopa:
Cc @shrishrirang
There is fix rolling out as part of ANT78 that allows setting java version and does not require setting web container. Should in prod over next few days.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -97,21 +103,29 @@ protected void createFunctionApp() throws Exception {
...
         final Update update = app.update();
+        // Check Java Version of Server
+        final JavaVersion serverJavaVersion = app.javaVersion();
+        if (serverJavaVersion.toString().startsWith("1.8")) {
+            info(String.format(SERVER_JAVA_VERSION, serverJavaVersion));
+        } else if (serverJavaVersion.equals(JavaVersion.OFF)) {
+            warning(SERVER_JAVA_VERSION_OFF);
+            update.withJavaVersion(JavaVersion.JAVA_8_NEWEST).withWebContainer(WebContainer.TOMCAT_9_0_NEWEST);
+        } else {
+            warning(SERVER_JAVA_VERSION_INCORRECT);
```
andxu: what's the requirement of java version?

Flanker32: For now we only support java 1.8

Flanker32:
@pragnagopa  Please help take a look, now we could set java version when update a function, but we still need to specify a webcontainer when create, and the default value is null in this pr.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -93,33 +103,44 @@ protected void createFunctionApp() throws Exception {
...
+        final JavaVersion serverJavaVersion = app.javaVersion();
+        if (serverJavaVersion.toString().matches(VALID_JAVA_VERSION_PATTERN)) {
+            info(String.format(FUNCTION_HOST_JAVA_VERSION, serverJavaVersion));
+        } else if (serverJavaVersion.equals(JavaVersion.OFF)) {
+            info(FUNCTION_HOST_JAVA_VERSION_OFF);
+            update.withJavaVersion(DEFAULT_JAVA_VERSION);
+        } else {
+            warning(FUNCTION_HOST_JAVA_VERSION_INCORRECT);
+            update.withJavaVersion(DEFAULT_JAVA_VERSION);
+        }
```
andxu: merge these with one brace and a conditional expr:
 else if (serverJavaVersion.equals(JavaVersion.OFF)) {
            info(FUNCTION_HOST_JAVA_VERSION_OFF);
            update.withJavaVersion(DEFAULT_JAVA_VERSION);
        } else {
            warning(FUNCTION_HOST_JAVA_VERSION_INCORRECT);
            update.withJavaVersion(DEFAULT_JAVA_VERSION);
        }

PR #498, opened 2018-11-15
Fix error message

SummerSun:


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/WarArtifactHandlerImpl.java
```diff
@@ -88,7 +88,7 @@ public void publish(final DeployTarget target) throws MojoExecutionException {
             }
         }
 
-        throw new MojoExecutionException(String.format(DEPLOY_FAILURE, retryCount));
+        throw new MojoExecutionException(String.format(DEPLOY_FAILURE, DEFAULT_MAX_RETRY_TIMES));
```
Flanker32: For the retry count is fixed, shall we just write the retry count in `DEPLOY_FAILURE `?

SummerSun: Give it another think, still need to use the retryCount.

PR #501, opened 2018-11-19
Check the region only when creating web app

SummerSun:
Remove the `createOrUpdateWebApp`, separate the create Web App logic and update Web App logic and move them into the `doExecute()` for better readability.

Only when creating a new Web App , the region is required.

### DIFF
--- a/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/parser/V2ConfigurationParser.java
+++ b/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/parser/V2ConfigurationParser.java
@@ -17,7 +17,6 @@ import com.microsoft.azure.maven.webapp.configuration.RuntimeSetting;
 import org.apache.maven.model.Resource;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.codehaus.plexus.util.StringUtils;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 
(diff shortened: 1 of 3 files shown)


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/parser/V2ConfigurationParser.java
```diff
@@ -51,9 +51,6 @@ protected OperatingSystemEnum getOs() throws MojoExecutionException {
     @Override
     protected Region getRegion() throws MojoExecutionException {
         final String region = mojo.getRegion();
-        if (StringUtils.isEmpty(region)) {
-            throw new MojoExecutionException("Please config the <region> in pom.xml.");
-        }
         if (Arrays.asList(Region.values()).contains(region)) {
```
Flanker32: !contains()

SummerSun:
test

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/parser/V2ConfigurationParser.java
```diff
@@ -51,7 +50,7 @@ protected OperatingSystemEnum getOs() throws MojoExecutionException {
     @Override
     protected Region getRegion() throws MojoExecutionException {
         final String region = mojo.getRegion();
-        if (!Arrays.asList(Region.values()).contains(region)) {
+        if (Region.findByLabelOrName(region) == null) {
```
Flanker32: What will happen if region is not config in pom here? will it also throw region not support Exception?

SummerSun: My fault. That case should be taken care of.

Issue #510, opened 2018-11-22
Deploy to a new web app with an existing app service plan, users could not update the pricing tier


sallyur:
### Plugin name and version
`Maven plugin for azure web apps: 1.5.1`
### Plugin configuration in your `pom.xml`
Take V2 as example:
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.ms.webapp</groupId>
  <artifactId>webapp0</artifactId>
  <packaging>war</packaging>
  <version>1.0-SNAPSHOT</version>
  <name>webapp0 Maven Webapp</name>
  <url>http://maven.apache.org</url>
  <dependencies>
    <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>3.8.1</version>
      <scope>test</scope>
    </dependency>
  </dependencies>

  <pluginRepositories>
    <pluginRepository>
      <id>commicrosoftazure-2715</id>
      <name>commicrosoftazure-2715</name>
      <url>https://oss.sonatype.org/content/repositories/commicrosoftazure-2715/</url>
    </pluginRepository>
  </pluginRepositories> 

  <build>
  <finalName>webapp0</finalName>
    <plugins>
      <plugin>
        <groupId>com.microsoft.azure</groupId>
        <artifactId>azure-webapp-maven-plugin</artifactId>
        <version>1.5.1</version>

        <configuration>
          <authentication>
            <serverId>azure-auth</serverId>
          </authentication>
          <resourceGroup>resource-Group-Name</resourceGroup>
          <appName>new-App-Name</appName>

          <!-- Deploy Web App to an existing Service Plan which pricing tier is not S3 -->
          <appServicePlanResourceGroup>your-existing-plan-group</appServicePlanResourceGroup>
          <appServicePlanName>your-existing-resource-plan-name</appServicePlanName>
          <region>northeurope</region>

          <pricingTier>S3</pricingTier>

          <schemaVersion>v2</schemaVersion>

          <runtime>
            <os>windows</os>
            <javaVersion>1.8</javaVersion>
            <webContainer>tomcat 8.5</webContainer>
          </runtime>

          <deployment>
            <resources>
              <resource>
[...]

PR #511, opened 2018-11-22
[ Hold ]Support to update the pricing tier of existing app service plan

SummerSun:
To address #510 
There aren't so many properties that could be updated once a Web App or App Service Plan is created. But the pricing tier is the one which could be updated. We should support this from our side.

- Remove `PricingTierEnum` and get the configuration as String to avoid multiple null checking and refactor the error message when users specify invalid pricing tier
- Support to update the service plan before creating or updating the web app if users specify an existing app service plan


Flanker32:
This has been done in #650

PR #516, opened 2018-11-28
Update web app samples

SummerSun:
Visit here for Markdown format: https://github.com/SummerSun/azure-maven-plugins/blob/r/docs/web-app-samples.md 

Correct the config `slotName` to `name` #519


SummerSun:
@jdneo 

It's a good idea. We currently group the configuration by platform and in the samples, the v1 goes first. We could discuss how to group the configuration by schema and make the v2 schema version at the beginning of the list.

PR #540, opened 2019-01-02
Webapp config

Flanker32:
Add config mojo to webapp maven plugin, it will generate new configuration when no configuration was found and it allow user to modify configurations.

Config includes three parts
* Webapp(appname, resourcegroup, pricingTier, region )
* Runtime(os, container and docker configuration)
* Slot(slot name and configurationSource)

When create new configuration, it will ask user for runtime attributes and create new configuration, when configuration was found, it will ask user which part to config and config correspond attributes.


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAppServiceMojo.java
```diff
@@ -25,13 +25,13 @@
     /**
      * Resource group of App Service. It will be created if it doesn't exist.
      */
-    @Parameter(property = "resourceGroup", required = true)
+    @Parameter(property = "resourceGroup",required = false)
```
andxu: don;t remove spaces

Flanker32:
Pr too large, split it into small one

PR #542, opened 2019-01-02
Add serializer to serialize webapp configurations to pom

Flanker32:
Add serializer to serialize webapp configurations to pom, for webapp:config


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/WebAppConfiguration.java
```diff
@@ -148,6 +154,40 @@ public MavenSession getSession() {
...
+
+    public DeploymentSlotSetting getDeploymentSlotSetting() {
+        return deploymentSlotSetting;
+    }
+
+    public String getSchemaVersion() {
+        return schemaVersion;
+    }
+
+    public Builder getBuilderFromConfiguration(){
```
jdneo: Where is this method being called?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/WebAppConfiguration.java
```diff
@@ -148,6 +154,40 @@ public MavenSession getSession() {
     public MavenResourcesFiltering getFiltering() {
         return filtering;
     }
+
+    public DeploymentSlotSetting getDeploymentSlotSetting() {
+        return deploymentSlotSetting;
+    }
+
+    public String getSchemaVersion() {
```
jdneo: Where is this method called?

Issue #550, opened 2019-01-10
When the pom.xml is incomplete, the error message is not accurate


sallyur:
The issue can repro when V1 convert to V2  or config V2
### Plugin name and version
```
Maven plugin for azure webapps
ID: 2832
Version: 1.5.2
```
### Plugin configuration in your `pom.xml`
Take V2 as example:
```
<?xml version="1.0" encoding="UTF-8"?>

<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">  
  <modelVersion>4.0.0</modelVersion>  
  <groupId>com.ms.webapp</groupId>  
  <artifactId>webapp0</artifactId>  
  <packaging>war</packaging>  
  <version>1.0-SNAPSHOT</version>  
  <name>webapp0 Maven Webapp</name>  
  <url>http://maven.apache.org</url>  
  <dependencies> 
    <dependency> 
      <groupId>junit</groupId>  
      <artifactId>junit</artifactId>  
      <version>3.8.1</version>  
      <scope>test</scope> 
    </dependency> 
  </dependencies>  
  <pluginRepositories> 
    <pluginRepository> 
      <id>commicrosoftazure-2832</id>  
      <name>commicrosoftazure-2832</name>  
      <url>https://oss.sonatype.org/content/repositories/commicrosoftazure-2832/</url> 
    </pluginRepository> 
  </pluginRepositories>  
  <build> 
    <finalName>webapp0</finalName>  
    <plugins> 
      <plugin> 
        <groupId>com.microsoft.azure</groupId>  
        <artifactId>azure-webapp-maven-plugin</artifactId>  
        <!-- check Maven Central for the latest version -->  
        <version>1.5.2</version>  
        <configuration>
          <schemaVersion>V2</schemaVersion>
          <resourceGroup>webapp01</resourceGroup>
          <appName>webapp01</appName>
          <region>westus2</region>
          <pricingTier>S1</pricingTier>
        
          <deployment>
            <resources>
              <resource>
                <directory>${project.basedir}/target</directory>
                <includes>
                  <include>*.war</include>
                </includes>
              </resource>
            </resources>
          </deployment>
        </configuration>
      </plugin> 
    </plugins> 
  </build> 
</project>
```

### Expected behavior
[...]

Issue #549, opened 2019-01-10
Inconsistent behavior during the execution of config


sallyur:
### Plugin name and version
```
Maven plugin for azure webapps
ID: 2832
Version: 1.5.2
```
### Plugin configuration in your `pom.xml`
```
<?xml version="1.0" encoding="UTF-8"?>

<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">  
  <modelVersion>4.0.0</modelVersion>  
  <groupId>com.ms.webapp</groupId>  
  <artifactId>webapp0</artifactId>  
  <packaging>war</packaging>  
  <version>1.0-SNAPSHOT</version>  
  <name>webapp0 Maven Webapp</name>  
  <url>http://maven.apache.org</url>  
  <dependencies> 
    <dependency> 
      <groupId>junit</groupId>  
      <artifactId>junit</artifactId>  
      <version>3.8.1</version>  
      <scope>test</scope> 
    </dependency> 
  </dependencies>  
  <pluginRepositories> 
    <pluginRepository> 
      <id>commicrosoftazure-2832</id>  
      <name>commicrosoftazure-2832</name>  
      <url>https://oss.sonatype.org/content/repositories/commicrosoftazure-2832/</url> 
    </pluginRepository> 
  </pluginRepositories>  
  <build> 
    <finalName>webapp0</finalName>  
    <plugins> 
      <plugin> 
        <groupId>com.microsoft.azure</groupId>  
        <artifactId>azure-webapp-maven-plugin</artifactId>  
        <!-- check Maven Central for the latest version -->  
        <version>1.5.2</version>  
        <configuration>
          <schemaVersion>V2</schemaVersion>
          <resourceGroup>webapp01</resourceGroup>
          <appName>webapp01</appName>
          <region>westus2</region>
          <pricingTier>S1</pricingTier>
          <runtime>
            <os>linux</os>
            <javaVersion>jre8</javaVersion>
            <webContainer>wildfly 14</webContainer>
          </runtime>
          <deployment>
            <resources>
              <resource>
                <directory>${project.basedir}/target</directory>
                <includes>
                  <include>*.war</include>
                </includes>
              </resource>
            </resources>
          </deployment>
        </configuration>
      </plugin> 
    </plugins> 
[...]

Issue #548, opened 2019-01-09
After run azure-webapp:config, there is no <deployment> element in pom.xml


sallyur:
### Plugin name and version
```
Maven plugin for azure webapps
ID: 2832
Version: 1.5.2
```

### Plugin configuration in your `pom.xml`

```
<?xml version="1.0" encoding="UTF-8"?>

<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">  
  <modelVersion>4.0.0</modelVersion>  
  <groupId>com.ms.webapp</groupId>  
  <artifactId>webapp0</artifactId>  
  <packaging>war</packaging>  
  <version>1.0-SNAPSHOT</version>  
  <name>webapp0 Maven Webapp</name>  
  <url>http://maven.apache.org</url>  
  <dependencies> 
    <dependency> 
      <groupId>junit</groupId>  
      <artifactId>junit</artifactId>  
      <version>3.8.1</version>  
      <scope>test</scope> 
    </dependency> 
  </dependencies>  
  <pluginRepositories> 
    <pluginRepository> 
      <id>commicrosoftazure-2832</id>  
      <name>commicrosoftazure-2832</name>  
      <url>https://oss.sonatype.org/content/repositories/commicrosoftazure-2832/</url> 
    </pluginRepository> 
  </pluginRepositories>  
  <build> 
    <finalName>webapp0</finalName>  
    <plugins> 
      <plugin> 
        <groupId>com.microsoft.azure</groupId>  
        <artifactId>azure-webapp-maven-plugin</artifactId>  
        <version>1.5.2</version>  
        
      </plugin> 
    </plugins> 
  </build> 
</project>

```
### Expected behavior
There should have deployment element in pom.xml 
### Actual behavior
There is no deployment element in pom.xml 
![image](https://user-images.githubusercontent.com/28860891/50895213-68c95c00-1440-11e9-9466-a9377fae9db8.png)

### Steps to reproduce the problem

1. Prepare a webapp maven project with neweast maven plugin version
2. Set the pom file as above
3. Run "mvn clean package"
4. Run "mvn azure-webapp:config"
5. Os and Runtime Stack can choose any value
6. The confirm step Input "N" 
7. The next steps can all set to default until the end

[...]

PR #552, opened 2019-01-10
Fix NPE issue and default values

Flanker32:
Fix npe when configuration is not complete. #550 #551
Fix init deployment issue #548
Fix default values #553 #549

### DIFF
--- a/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
+++ b/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
@@ -180,7 +180,7 @@ public class ConfigMojo extends AbstractWebAppMojo {
             defaultResourceGroup,
             NOT_EMPTY_REGEX, null, null);
 
-        final String defaultRegion = getDefaultRigon(configuration.getRegion());
+        final String defaultRegion = getDefaultRegion(configuration.getRegion());
         final String region = queryer.assureInputFromUser("region", defaultRegion, NOT_EMPTY_REGEX,
             null, null);
 
@@ -319,7 +319,7 @@ public class ConfigMojo extends AbstractWebAppMojo {
         return StringUtils.isNotEmpty(defaultValue) ? defaultValue : fallBack;
     }
 
-    private String getDefaultRigon(Region region) {
+    private String getDefaultRegion(Region region) {
         return region != null ? region.toString() : DEFAULT_REGION.toString();
     }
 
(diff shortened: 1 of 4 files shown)


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
```diff
@@ -184,17 +175,20 @@ private WebAppConfiguration getWebAppConfiguration(WebAppConfiguration configura
             NOT_EMPTY_REGEX, null, null);
 
         final String defaultRegion = configuration.region != null ? configuration.region.name() :
```
andxu: add EUROPE_WEST as a constant default value in configuration, provide a getRegion method returning with the failback-to-default code.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
```diff
@@ -40,10 +40,10 @@
     public static final String NOT_EMPTY_REGEX = "[\\s\\S]+";
     public static final String BOOLEAN_REGEX = "[YyNn]";
 
-    public static final String CONFIGURATION_NOT_FOUND = "Configuration not found, init new configuration.";
-    public static final String CONVERT_TO_V2SCHEMA = "Converting configuration to V2 schema";
     public static final String SAVING_TO_POM = "Saving configuration to pom.";
-    public static final String EXITING = "Config only support V2 schema, exiting";
+    public static final String CONFIG_ONLY_SUPPORT_V2 = "Config only support V2 schema";
+    public static final String CHANGE_OS_WARNING = "The plugin may not work if you change the os of an existing " +
```
jdneo: Why we provide this warning here?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
```diff
@@ -104,30 +92,33 @@ protected boolean confirmConfiguration(WebAppConfiguration configuration) throws
...
-            case Docker:
-                System.out.println("Image : " + configuration.getImage());
-                System.out.println("ServerId : " + configuration.getServerId());
-                System.out.println("RegistryUrl : " + configuration.getRegistryUrl());
-                break;
-            default:
-                throw new MojoExecutionException("The value of <os> is unknown.");
+
+        if (configuration.getOs() == null) {
+            System.out.println("No runtime configuration");
```
jdneo: What information do we hope the users to get here?

If the output is only `No runtime configuration`. Then this is much more like a log in debug mode.

Flanker32: For we support user didn't config runtime and deploy, and for now we don't know whether user deploy to an existing webapp or just create a new one, so add this information to warn user they didn't complete the runtime if they plan to deploy a new webapp.

Issue #561, opened 2019-01-15
[Functions][Regression] Cardinality is missing for EventHubTrigger


pragnagopa:
### Plugin name and version
azure.functions.maven.plugin
1.2.1
### Expected behavior
EventHubTrigger with following signature:
```
 @EventHubTrigger(name = "messages", eventHubName = "test-input-java", connection = "AzureWebJobsEventHubSender", dataType = "string", cardinality = Cardinality.MANY) String[] messages,
```
should generate function.json with Cardinality Many
```
{
    "type" : "eventHubTrigger",
    "name" : "messages",
    "direction" : "in",
    "dataType" : "string",
    "consumerGroup" : "$Default",
    "connection" : "AzureWebJobsEventHubSender",
    "cardinality" : "many",
    "eventHubName" : "test-input-java"
  }
```
### Actual behavior
in Version 1.2.1 , Cardinality is missing

pragnagopa:
@Flanker32 / @jdneo - This is a regression and needs to be addressed ASAP.

Flanker32:
@pragnagopa  For now we write values differs from its default value to function.json as we don't know whether the value is set by users, or is there anyway for maven tooling to find whether an attribute is required in function.json like schema or annotation?

Flanker32:
FIxed in #562

PR #562, opened 2019-01-16
Always write cardinality to function.json

Flanker32:
Always write cardinality to function.json, for issue #561

### DIFF
--- a/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/Binding.java
+++ b/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/Binding.java
@@ -10,7 +10,9 @@ import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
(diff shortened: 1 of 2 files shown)


jdneo:
commit message typo: Alawys

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/Binding.java
```diff
@@ -24,6 +26,13 @@
 
     protected Map<String, Object> bindingAttributes = new HashMap<>();
 
+    protected static Map<BindingEnum, List<String>> requiredAttributeMap = new HashMap<>();
+
+    static {
+        //initialize required attributes, which will save to function.json even if it equals to its default value
```
andxu: be saved

Flanker32:
@pragnagopa  Test case has been added, one question is maven plugin will read the property of annotations and write it to function.json directly, and the value will be like `"cardinality" : "MANY"`, will it work as well ?

pragnagopa:
Yes. "cardinality" : "MANY" works

Issue #559, opened 2019-01-14
[Functions] Add support for CustomBinding


pragnagopa:
PR https://github.com/Azure/azure-functions-java-library/pull/76 adds CustomBinding annotation.
Plugin needs to be updated to generate function.json for the customBinding annotation.
Sample code:
```
@CustomBinding(direction = "in", name = "customInput", type = "customTrigger") String customTriggerInput
```
should generated function.json
```
"type" : "customTrigger",
"direction" : "in",
"name" : "customInput",
 ```

**Comment by pragnagopa** [comment]
@Flanker32 / @jdneo - This is a tracking item. Will update the issue once snapshot version of library is released.

**Comment by pragnagopa** [comment]
azure-functions-java-library version 1.3.0-SNAPSHOT is now available in snapshot repository. 
Following function
```java
public void CustomBinding_Valid(@HttpTrigger(name = "req", methods = { HttpMethod.GET,
      HttpMethod.POST }, authLevel = AuthorizationLevel.FUNCTION) HttpRequestMessage<String> request,
      @TestCustomBinding(index = "testIndex", path = "testPath") String customInput) {
  }
```
TestCustomBinding is defined
```
@Target(ElementType.PARAMETER)
@Retention(RetentionPolicy.RUNTIME)
@CustomBinding(direction = "in", name = "message", type = "customBinding")
public @interface TestCustomBinding {
   String index();
   String path();
}
```

should generate
```
"type" : "customBinding",
"direction" : "in",
"name" : "message",
"index " : "testIndex",
"path" : "testPath"
```

Flanker32:
@pragnagopa  i just find `CustomBinding` in function java lib with `name`,'type' and 'direction', i didn't find `TestCustomBinding `, is there anything wrong with the version?

pragnagopa:
TestCustomBinding Is user defined.

pragnagopa:
Here is sample code for invoking methods on a user defined annotation
```java
for (Method annotationMethod : annotation.annotationType().getDeclaredMethods()) {
						try {
							if (Modifier.isPublic(annotationMethod.getModifiers())) {
								Object invokeResult = annotationMethod.invoke(annotation);
								System.out.println(annotationMethod.getName() + "=" + invokeResult);
							}
						} catch (Exception ex) {
							// ignore
						}
					}
```

Generate key value pairs in function.json. No validation required.

PR #570, opened 2019-01-25
Add support for custom binding

Flanker32:
Add support for custom binding, for issue #559


azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/CustomBinding.java
```diff
@@ -0,0 +1,32 @@
...
+ * license information.
+ */
+
+package com.microsoft.azure.maven.function.bindings;
+
+import java.lang.annotation.Annotation;
+
+public class CustomBinding extends Binding {
+
+    private com.microsoft.azure.functions.annotation.CustomBinding customBindingAnnotation;
```
andxu: what's the reason for using full package name here? com.microsoft.azure.functions.annotation.CustomBinding

Flanker32: This is the annotation class from azure function java library, in order not to mix it with CustomBinding Object, use the full name.

Flanker32:
@pragnagopa  Sorry for reply late, cases has been updated

azure-functions-maven-plugin/src/test/java/com/microsoft/azure/maven/function/handlers/AnnotationHandlerImplTest.java
```diff
@@ -87,7 +94,24 @@
...
+
+        @FunctionName(CUSTOM_BINDING_FUNCTION)
+        public void customBindingMethod(
+            @TestCustomBinding(index = "testIndex", path = "testPath") String customTriggerInput) {
+        }
+
+        @Target(ElementType.PARAMETER)
+        @Retention(RetentionPolicy.RUNTIME)
+        @CustomBinding(direction = "in", name = "message", type = "customBinding")
+        public @interface TestCustomBinding {
```
pragnagopa: Using CustomBinding by itself is supported as well. Please add test.

pragnagopa:
@Flanker32 - we should also support
```java
@Target(ElementType.PARAMETER)
@Retention(RetentionPolicy.RUNTIME)
@CustomBinding(direction = "in", name = "", type = "customBinding")
public @interface TestCustomBinding {
   String index();
   String path();
   String name();
}
```
In this case, name() on TestCustomBinding should override name given in customBinding

jdneo:
@pragnagopa It's a good idea that using CustomBinding to extends the binding model. But may I ask in what kind of scenario, the user needs to override the `name`, `direction` or `type` fields defined in `@CustomBinding`. So far, I didn't see the necessity for this kind of usage.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/Binding.java
```diff
@@ -100,4 +100,21 @@ public String toString() {
                 .append(" ]")
                 .toString();
     }
+
+    protected String getDirectionFromAttribute() {
+        if (this.direction != null) {
```
Flanker32: We may split this function, do the null check of direction in `getDirection` and only get direction from attribute map here.

jdneo:
@pragnagopa Could you please take a look?

anthonychu:
> @pragnagopa It's a good idea that using CustomBinding to extends the binding model. But may I ask in what kind of scenario, the user needs to override the `name`, `direction` or `type` fields defined in `@CustomBinding`. So far, I didn't see the necessity for this kind of usage.

@jdneo The main scenario is to use `CustomBinding` to define annotations to be used in multiple places (therefore it is useful to control the name where the annotation is used). We've [been asked](https://github.com/Azure/azure-functions-java-library/pull/78#pullrequestreview-203504473) to use `CustomBinding` to build annotations for SignalR Service. It's not possible to do this without support for overriding the name (overriding `direction` and `type` are not necessary), so this is currently a blocker for us.

Here's an example:

```java
@Target({ElementType.PARAMETER, ElementType.METHOD})
@Retention(RetentionPolicy.RUNTIME)
@CustomBinding(direction = "out", name = "", type = "SignalR")
public @interface SignalR {
   String name();
   String hubName();
}
```

Usage:

```java
@FunctionName("messages")
public void sendMessage(
        @HttpTrigger(
            name = "req", 
            methods = { HttpMethod.POST },
            authLevel = AuthorizationLevel.ANONYMOUS) 
            HttpRequestMessage<Object> req,
        @SignalR(name = "chatMessage", hubName = "chat") OutputBinding<SignalRMessage> chatMessage,
        @SignalR(name = "notificationMessage", hubName = "notification") OutputBinding<SignalRMessage> notificationMessage) {

    chatMessage.setValue(/* a chat message */);
    notificationMessage.setValue(/* a notification message */);
}
```

/cc @JialinXin @chenkennt @sffamily

jdneo:
@anthonychu Thank you for clarifying. Get your points now!

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/BindingEnum.java
```diff
@@ -32,7 +32,8 @@
     TableOutput("table", Direction.OUT, true),
     TimerTrigger("timerTrigger", Direction.IN),
     TwilioSmsOutput("twilioSms", Direction.OUT),
-    CustomBinding("customBinding", null);
+    CustomBinding("customBinding", null),
+    UnknownBinding(null, null);
```
Flanker32: I think ExtendedCustomBindingEnum could be better

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/Binding.java
```diff
@@ -101,20 +102,22 @@ public String toString() {
                 .toString();
     }
 
-    protected String getDirectionFromAttribute() {
-        if (this.direction != null) {
-            return direction.toString();
+    protected void addProperties(List<String> requiredAttributes, Object value, Method method) {
```
Flanker32: I think we don't need to pass requiredAttributes as it is an member properties of Binding class

jdneo:
Currently, the plugin should be able to resolve `CustomBinding` and the annotations which are extended from `CustomBinding`. While from the implementation, we found that the logic needs to be refactored.

Considering Java GA is coming soon, we decide to put the refactoring work after GA. Tracking issue: #583

azure-functions-maven-plugin/src/test/java/com/microsoft/azure/maven/function/handlers/AnnotationHandlerImplTest.java
```diff
@@ -87,7 +96,33 @@
...
+        }
+
+        @FunctionName(CUSTOM_BINDING_FUNCTION)
+        public void customBindingMethod(
+                @CustomBinding(name = "input", type = "customBinding", direction = "in") String input) {
+        }
+
+        @FunctionName(EXTENDING_CUSTOM_BINDING_FUNCTION)
+        public void extendingcustomBindingMethod(
+            @TestCustomBinding(direction = "out", name = "extendingCustomBinding", path = "testPath")
```
pragnagopa: direction cannot be overridden as the annotation author decides the direction for a given binding type.

azure-functions-maven-plugin/src/test/java/com/microsoft/azure/maven/function/handlers/AnnotationHandlerImplTest.java
```diff
@@ -87,7 +96,33 @@
...
+        @Retention(RetentionPolicy.RUNTIME)
+        @CustomBinding(direction = "in", name = "message", type = "customBinding")
+        public @interface TestCustomBinding {
+            String direction();
+            String path();
+            String name();
+        }
+
+        @FunctionName(CUSTOM_BINDING_FUNCTION)
+        public void customBindingMethod(
```
pragnagopa: Add one more test where TestCustomBinding does not define a name.
```java
@FunctionName(EXTENDING_CUSTOM_BINDING_FUNCTION)
        public void extendingcustomBindingMethod2(
            @TestCustomBinding(direction = "out",  path = "testPath")
                    String customTriggerInput) {
        }
```
In this case, name should be taken from customBinding: message

anthonychu:
I just tested this out as well. It works great. Thanks for the quick work on this @jdneo

Updated this sample to use it:
https://github.com/Azure-Samples/signalr-service-quickstart-serverless-chat/blob/antchu/add-java/src/chat/java/src/main/java/com/function/Functions.java

azure-functions-maven-plugin/src/test/java/com/microsoft/azure/maven/function/handlers/AnnotationHandlerImplTest.java
jdneo: Does that mean only name can be overridden?

jdneo:
@anthonychu Glad to know it works. I just make some tiny changes to the code to address the comment. Would you mind to try it again to make sure things go well?

@pragnagopa Regarding this PR, We found some problems. Could you please take a look?
- https://github.com/Azure/azure-functions-java-library/issues/83
- https://github.com/Azure/azure-functions-java-library/issues/69

Issue #568, opened 2019-01-24
The terminal window is no response when config V1 project


sallyur:
### Plugin name and version
```
Maven plugin for Azure webapps
ID: commicrosoftazure-2862
Version: 1.5.3
```
### Plugin configuration in your `pom.xml`
```
<?xml version="1.0" encoding="UTF-8"?>

<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">  
  <modelVersion>4.0.0</modelVersion>  
  <groupId>com.ms.webapp</groupId>  
  <artifactId>webapp0</artifactId>  
  <packaging>war</packaging>  
  <version>1.0-SNAPSHOT</version>  
  <name>webapp0 Maven Webapp</name>  
  <url>http://maven.apache.org</url>  
  <dependencies> 
    <dependency> 
      <groupId>junit</groupId>  
      <artifactId>junit</artifactId>  
      <version>3.8.1</version>  
      <scope>test</scope> 
    </dependency> 
  </dependencies>  
  <pluginRepositories> 
    <pluginRepository> 
      <id>commicrosoftazure-2862</id>  
      <name>commicrosoftazure-2862</name>  
      <url>https://oss.sonatype.org/content/repositories/commicrosoftazure-2862/</url> 
    </pluginRepository> 
  </pluginRepositories>  
  <build> 
    <finalName>webapp0</finalName>  
      <plugins>
         <plugin>
            <groupId>com.microsoft.azure</groupId>
            <artifactId>azure-webapp-maven-plugin</artifactId>
             <!-- check Maven Central for the latest version -->
            <version>1.5.3</version>
            <configuration>
               <!-- Referencing <serverId> in Maven's settings.xml to authenticate with Azure -->
               <authentication>
                 <serverId>azure-auth</serverId>
               </authentication>
               
               <!-- Web App information -->
               <resourceGroup>your-resource-group</resourceGroup>
               <appName>your-app-name</appName>
               <region>northeurope</region>
               <pricingTier>S3</pricingTier>

               <!-- Java Runtime Stack for Web App on Windows-->     
               <javaVersion>1.8</javaVersion>      
               <javaWebContainer>tomcat 8.5</javaWebContainer>
               
               <!-- WAR deplotyment type settings -->
[...]

Issue #341, opened 2018-08-13
Maven command will be hanged on MacOS


jdneo:
Platform Information
Operation System: MacOS 10.13.5
Java Version: Java version: 1.8.0_162, vendor: Oracle Corporation
Maven Version: 3.5.0
AI-Core Version: 2.1.2

> Note: This issue seems cannot be reproduced on Windows, but can be stable repro on Mac

Seems the AI will send the telemetry asynchronously. So when the maven goal finished, maven dispose its classloader, but the AI is still doing some tasks. Then NoClassDefFoundError happens. Besides the process will be hanged.

This issue is also tracked by: https://github.com/Microsoft/ApplicationInsights-Java/issues/416

Flanker32:
Fixed in #576

PR #576, opened 2019-01-31
Fix AI SDK may step into endless loop when maven goal execute too quick

Flanker32:
Shutdown JVM when AI SDK step into endless loop, for issue #341 #568

### DIFF
--- a/azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
+++ b/azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
@@ -6,6 +6,7 @@
 
 package com.microsoft.azure.maven;
 
+import com.microsoft.applicationinsights.internal.channel.common.ApacheSenderFactory;
 import com.microsoft.azure.management.Azure;
 import com.microsoft.azure.maven.auth.AuthConfiguration;
 import com.microsoft.azure.maven.auth.AuthenticationSetting;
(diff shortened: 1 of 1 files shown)


jdneo:
Another workaround is calling `ApacheSenderFactory.INSTANCE.create().close();` in the Main thread. But still may lose telemetry... I think we need more investigation about this issue.

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
```diff
@@ -406,6 +406,14 @@ private void updateConfigurationFile(Properties prop) {
...
         public void uncaughtException(Thread t, Throwable e) {
             debug("uncaughtException: " + e);
+            // Work around for Application Insights Java SDK:
+            // Related issues: https://github.com/Microsoft/azure-maven-plugins/issues/341
+            // When maven goal executes too quick, AI SDK may not fully initialized and will step into endless loop
+            // Refer here for detail codes: https://github.com/Microsoft/ApplicationInsights-Java/blob/master/core/src
+            // /main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java#L103
+            // As ShutdownHook(SDKShutdownActivity.ShutdownAction) will call this method and we have no other way to
+            // stop a shutdown hook thread, call Runtime.halt() here and this should be removed once AI issue fixed.
+            Runtime.getRuntime().halt(0);
```
andxu: Cannot call   Runtime.getRuntime().halt(0)  when arbitrary exception occurs.

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
```diff
@@ -311,6 +312,12 @@ public void execute() throws MojoExecutionException {
             }
         } catch (Exception e) {
             handleException(e);
+        }finally{
```
andxu: need a space.

PR #585, opened 2019-02-26
Fix checkstyle dtd schema error

jdneo:
See:  https://github.com/checkstyle/checkstyle/issues/1571

### DIFF
--- a/build-tools/src/main/resources/checkstyle/checkstyle-suppressions.xml
+++ b/build-tools/src/main/resources/checkstyle/checkstyle-suppressions.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE suppressions PUBLIC
-    "-//Checkstyle//DTD Suppressions 1.3//EN"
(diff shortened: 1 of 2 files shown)


build-tools/src/main/resources/checkstyle/checkstyle-suppressions.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE suppressions PUBLIC
-    "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
-    "http://www.puppycrawl.com/dtds/suppressions_1_0.dtd">
+    "-//Checkstyle//DTD Suppressions 1.3//EN"
+    "http://checkstyle.sourceforge.net/dtds/suppressions_1_0.dtd">
```
Flanker32: Seems 1.1 is the latest version, or we keep use 1.0?
<!DOCTYPE suppressions PUBLIC
    "-//Checkstyle//DTD Suppressions 1.1//EN"
    "http://checkstyle.sourceforge.net/dtds/suppressions_1_1.dtd">

Issue #571, opened 2019-01-25
[Functions] Print function URLs on publish


asavaritayal:
(This has been a common piece of feedback, especially for HTTP triggered functions. )

**Problem statement:** 
Once the user deploys their Java project using the `azure-functions:deploy` target, they have no clue how to call the function without navigating to the portal to "Get function URL <>". This requires prior knowledge of the portal UI. Also creates an additional step in the E2E experience. 

**Options:** 
The Core Tools recently introduced an update to publish the function URLs to the CLI once the deployment completes: 

![image](https://user-images.githubusercontent.com/8435148/51772297-7ddcf680-20a0-11e9-9535-6755b80c8a70.png)

It'll be great to have a similar experience in the Maven CLI as well. 

/cc @hexiaokai @selvasingh

Flanker32:
@asavaritayal 
Yes, it do values and i will try to implement it with HTTP trigers.

Flanker32:
@asavaritayal Sorry for reply late, this was implemented in pr #596, for now we use azure java sdk to implement it as list function of a function app is not supported with mgmt sdk.

asavaritayal:
@Flanker32 thanks for the update. Know that @mathewc and @divyagandhi01 made some improvements on the ARM layer to enable the same scenario. Calling the APIs directly may be an option if the management SDK doesn't work out.

ggailey777:
Just an FYI...Visual Studio Code extension doesn't do this automatically, but they provide UI to get the URL of individual HTTP endpoints, which is very useful when function keys are used. I have requested a similar feature for the VS Functions tools.

Flanker32:
Done in https://github.com/microsoft/azure-maven-plugins/pull/1106

PR #596, opened 2019-03-15
[ Hold ]Show url of http trigger with authLevel ANONYMOUS after deployment

Flanker32:
Show url of http trigger with authLevel ANONYMOUS after deployment, for issue #571 

Known issue:
Maven will show warning when execute function maven goal.
`
[WARNING] The POM for com.microsoft.azure.appservice.v2016_08_01:azure-mgmt-appservice:jar:1.0.0-beta-1 is invalid, transitive dependencies (if any) will not be available, enable debug logging for more details
`

### DIFF
--- a/azure-functions-maven-plugin/pom.xml
+++ b/azure-functions-maven-plugin/pom.xml
@@ -199,6 +199,16 @@
             <version>${azure.maven-plugin-lib.version}</version>
         </dependency>
         <dependency>
+            <groupId>com.microsoft.azure.appservice.v2016_08_01</groupId>
+            <artifactId>azure-mgmt-appservice</artifactId>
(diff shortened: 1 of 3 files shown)


asavaritayal:
Does this cover only anonymous functions or other authentication levels as well? Just wanted to check since the original issue targeted both scenarios, later being the more widespread amongst users.

Flanker32:
> Does this cover only anonymous functions or other authentication levels as well? Just wanted to check since the original issue targeted both scenarios, later being the more widespread amongst users.

@asavaritayal  For now it only contains anonymous HTTP-trigger functions, as other functions need auth with its function key. It's possible to get the keys as mgmt sdk provide related methods.but there may have safety issue to print keys directly.

PR #645, opened 2019-05-06
Use NetworkInterface get mac address when ifconfig/getmac fails

Flanker32:
Use NetworkInterface get mac address when `ifconfig`/`getmac` fails.


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/telemetry/GetHashMac.java
```diff
@@ -13,40 +13,39 @@
...
-    public static boolean isValidHashMacFormat(String hashMac) {
-        if (hashMac == null || hashMac.isEmpty()) {
-            return false;
-        }
-
-        final Pattern hashedMacPattern = Pattern.compile(HASHED_MAC_REGEX);
-        final Matcher matcher = hashedMacPattern.matcher(hashMac);
-        return matcher.matches();
-    }
+    public static final Pattern MAC_PATTERN = Pattern.compile(MAC_REGEX);
```
andxu: can MAC_PATTERN  be private?

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/telemetry/GetHashMac.java
```diff
@@ -85,6 +88,29 @@ private static String getRawMac() {
         return ret;
     }
 
+    private static String getRawMacWithoutIfconfig() {
+        final List<String> macSet = new ArrayList<>();
+        try {
+            final Enumeration<NetworkInterface> interfaces = NetworkInterface.getNetworkInterfaces();
```
andxu: log an event when ifconfig is not available(this code is called)

Flanker32: Done

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/telemetry/GetHashMac.java
Flanker32: Yes, they are used just here.

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/telemetry/GetHashMac.java
Flanker32: This method is called while Telemetry Proxy is initializing, so we can't send telemetries here.

PR #662, opened 2019-05-24
Remove F1 reference from pricing tiers

kyleburnsdev:
When configuring with F1, the plugin fails because it attempts to set AlwaysOn. Since AlwaysOn is a best practice to avoid horrible soft start times, updating the readme to avoid confusion

### DIFF
--- a/azure-webapp-maven-plugin/README.md
+++ b/azure-webapp-maven-plugin/README.md
@@ -244,10 +244,11 @@ Some valid regions are listed as below. All valid regions please refer to [Azure
 
 #### `<pricingTier>`
 Some valid pricing tiers are listed as below. All valid pricing tiers please refer to [Azure App Service Plan Pricing](https://azure.microsoft.com/en-us/pricing/details/app-service/).
+- `F1` (Windows Only)
 - `P1V2`
 - `P2V2`
 
-> Note: Pricing tier selected should support the 'AlwaysOn' setting for deployment to succeed
+> Note: Pricing tier selected should support the 'AlwaysOn' setting for deployment to succeed.
 
 ## Samples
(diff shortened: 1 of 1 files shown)


Flanker32:
@kyleburnsdev Thanks for your support. I tried to deploy a windows web app to Azure with maven plugin and it works well. And I could see upgrade to enable always on in portal. (Seems server side close always on automatically)
image
For Linux app service, I tried to close always on and the server returns `Free and Shared SKUs are not available for apps running on Linux`. So I think we could prompt users that F1 only support Windows app service instead of removing it, for users may want to try with a free one first.

kyleburnsdev:
> @kyleburnsdev Thanks for your support. I tried to deploy a windows web app to Azure with maven plugin and it works well. And I could see upgrade to enable always on in portal. (Seems server side close always on automatically)
> 
> For Linux app service, I tried to close always on and the server returns Free and Shared SKUs are not available for apps running on Linux. So I think we could prompt users that F1 only support Windows app service instead of removing it, for users may want to try with a free one first.

I updated to re-add F1, calling out that it is for Windows only

azure-webapp-maven-plugin/README.md
```diff
@@ -244,10 +244,12 @@ Some valid regions are listed as below. All valid regions please refer to [Azure
 
 #### `<pricingTier>`
 Some valid pricing tiers are listed as below. All valid pricing tiers please refer to [Azure App Service Plan Pricing](https://azure.microsoft.com/en-us/pricing/details/app-service/).
-- `F1`
+- `F1` (Windows Only)
 - `P1V2`
 - `P2V2`
 
+> Note: Pricing tier selected should support the 'AlwaysOn' setting for deployment to succeed.
```
Flanker32: I think we could prompt users that some pricing tier (like F1) didn't support Always on and may meet exceptions while deploying, as users don't know which pricing tier support `Always on`

Issue #660, opened 2019-05-23
webapp:config should generate version tag if <plugin> is absent, otherwise next deploy command will fail due to missing version tag.


andxu:
use this command `mvn com.microsoft.azure:azure-webapp-maven-plugin:1.6.0:config` on a new helloword sample.

Flanker32:
Fixed in #664

PR #664, opened 2019-05-28
Generate version tag if <plugin> is absent while `webapp:config`

Flanker32:
Generate version tag if <plugin> is absent while `webapp:config`, for issue #660

### DIFF
--- a/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/WebAppPomHandler.java
+++ b/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/WebAppPomHandler.java
@@ -29,7 +29,7 @@ public class WebAppPomHandler {
 
     public static final String PLUGIN_GROUP_ID = "com.microsoft.azure";
     public static final String PLUGIN_ARTIFACT_ID = "azure-webapp-maven-plugin";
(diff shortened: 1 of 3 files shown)


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/WebAppPomHandler.java
```diff
@@ -22,12 +22,15 @@
...
+import java.util.Properties;
 
 public class WebAppPomHandler {
 
-    public static final String PLUGIN_GROUPID = "com.microsoft.azure";
-    public static final String PLUGIN_ARTIFACTID = "azure-webapp-maven-plugin";
-    public static final String PLUGIN_XPATH = "/build/plugins/plugin[groupId=%s,artifactId=%s]";
+    public static final String PLUGIN_GROUP_ID = "com.microsoft.azure";
+    public static final String PLUGIN_ARTIFACT_ID = "azure-webapp-maven-plugin";
+    public static final String PROPERTIES_FILE = "plugin.properties";
```
andxu: PLUGIN_PROPERTIES_FILE

Flanker32: Fixed

Issue #683, opened 2019-06-21
Will prompt twice when I update service plan


EricZhangzf:
### Plugin name and version
azure-webapp-maven-plugin 1.7.0
### Plugin configuration in your `pom.xml`
```
<build>
   <finalName>webapp10</finalName>
   <plugins>
      <plugin>
         <groupId>com.microsoft.azure</groupId>
         <artifactId>azure-webapp-maven-plugin</artifactId>
         <version>1.7.0</version>
         <configuration>
            <resourceGroup>eric621gp</resourceGroup>
            <appName>eric-v1-war</appName>
            <appServicePlanResourceGroup>eric-621-war</appServicePlanResourceGroup>
            <appServicePlanName>eric-677war</appServicePlanName>
            <linuxRuntime>tomcat 9.0-jre8</linuxRuntime>
            <appSettings>
               <property>
                  <name>key</name>
                  <value>value</value>
               </property>
            </appSettings>
         </configuration>
      </plugin>
   </plugins>
</build>
```
### Expected behavior
It should be only prompt once.
### Actual behavior
String "Updating app service plan will prompt twice"
![image](https://user-images.githubusercontent.com/44559777/59910494-9a5d3f00-9444-11e9-8405-8746989d59c1.png)

### Steps to reproduce the problem
1. Deploy to a web app with creating a new plan;
2. After deployment, modify service plan name to another existing service plan.
3. Deploy again.

EricZhangzf:
Tested Pass.

Issue #682, opened 2019-06-21
There is no warning when webapp:config v1 configurations


bsaby:
### Plugin name and version
azure-webapp-maven-plugin 1.7.0
### Plugin configuration in your `pom.xml`
![image](https://user-images.githubusercontent.com/22470312/59913075-1443f700-944a-11e9-8a5d-89583a3189e0.png)

### Expected behavior
Show warning
![image](https://user-images.githubusercontent.com/22470312/59913258-6a189f00-944a-11e9-853c-ff374d43a373.png)

### Actual behavior
Can generate v2 configurations in pom file mixed with v1 configurations
![image](https://user-images.githubusercontent.com/22470312/59913265-6dac2600-944a-11e9-88e0-d6328ee01ca1.png)

### Steps to reproduce the problem
1. Create a web app project and add some v1 configurations
2. Run `mvn azure-webapp:config `

bsaby:
The test passed.

Issue #680, opened 2019-06-21
Cannot generate or update linux runtime


bsaby:
### Plugin name and version
azure-webapp-maven-plugin 1.7.0
### Plugin configuration in your `pom.xml`
BEFORE
```
<build>
    <finalName>webapp0</finalName>
</build> 
```
AFTER
```
<build> 
    <finalName>webapp0</finalName>  
    <plugins>
      <plugin>
        <groupId>com.microsoft.azure</groupId>
        <artifactId>azure-webapp-maven-plugin</artifactId>
        <version>1.7.0</version>
        <configuration>
          <schemaVersion>V2</schemaVersion>
          <resourceGroup>webapp0-1561105943027-rg</resourceGroup>
          <appName>webapp0-1561105943027</appName>
          <region>westeurope</region>
          <pricingTier>P1v2</pricingTier>
          <runtime>
            <os>linux</os>
          </runtime>
          <deployment>
            <resources>
              <resource>
                <directory>${project.basedir}/target</directory>
                <includes>
                  <include>*.war</include>
                </includes>
              </resource>
            </resources>
          </deployment>
        </configuration>
      </plugin>
    </plugins>
  </build>
```
### Expected behavior
Generate the complete Linux configuration in pom.xml
```
<runtime>
      <os>Linux</os>                 
      <javaVersion>jre8</javaVersion>
      <webContainer>tomcat 8.5</webContainer>
</runtime>
```
### Actual behavior
Only has os value
```
<runtime>
    <os>linux</os>
</runtime>
```
Same issue while updating Linux runtime by command `mvn azure-webapp:config`
### Steps to reproduce the problem
1. Generate a web app project
2. Generate configuration using `mvn com.microsoft.azure:azure-webapp-maven-plugin:1.7.0:config`, you can use all default value

bsaby:
Test passed.
Can generate and update Linux runtime.

PR #685, opened 2019-06-24
Fix issues in eng-game test

Flanker32:
Fix the method to get schema version, for issue #682 
Fix config can't update linux runtime node, for issue #680
Don't update app service plan with get method, for issue #683

### DIFF
--- a/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
+++ b/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
@@ -76,7 +76,7 @@ public class ConfigMojo extends AbstractWebAppMojo {
     }
 
     private boolean isV2Configuration(WebAppConfiguration configuration) {
-        return configuration == null || configuration.getSchemaVersion().equals(SchemaVersion.V2.toString());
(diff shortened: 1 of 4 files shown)


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
```diff
@@ -75,8 +75,8 @@ protected void doExecute() throws Exception {
         }
     }
 
-    private boolean isV1Configuration(WebAppConfiguration configuration) {
-        return configuration != null && configuration.getSchemaVersion().equals(SchemaVersion.V1.toString());
+    private boolean isV2Configuration(WebAppConfiguration configuration) {
+        return configuration == null || schemaVersion.equals(SchemaVersion.V2.toString());
```
andxu: configuration.getSchemaVersion().equals(SchemaVersion.V2.toString()) ?

Flanker32: Configuration in config mojo is generated by `V2NoValidationConfigurationParser`, which extend `V2ConfigurationParser` and will return `V2` when getSchemaVersion directly.

PR #693, opened 2019-07-08
Add the oauth/device login if credentials cannot be found.

andxu:
* When user doesn't install *Azure CLI* or he doesn't login using `az login`, the deploy command will fail reporting *Failed to authenticate with Azure. Please check your configuration*
 * With changes in this PR, *maven webapp&function plugin* will open a browser allowing user to choose an account to oauth login, and when user chooses the right account, the deployment will follow. If there is no browser(Eg: unix shell), it will give user a device login url and a code, both login methods have the similar behaviour with the command `az login`, user doesn't noeed to install *azure cli* at all. Be aware that no credentials are saved to local disk. If the user has multiple subscriptions he needs to set the subscription id as the following example:
 
 ```xml
 <configuration>
  <schemaVersion>V2</schemaVersion>
  <subscriptionId>685ba005-af8d-4b04-8f16-a7bf38b2eb5a</subscriptionId>
  ...
</configuration>
```

### CODE REVIEW

File: azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/auth/AzureAuthHelper.java [hunk 1]
```diff
@@ -160,10 +165,44 @@ protected Authenticated getAuthObj() {
...
+            if (cred == null) {
+                // fallback to device login
+                cred = AzureLoginHelper.deviceLogin(env);
+            }
+            // if all tries fails
+            if (cred == null) {
+                return null;
+            }
+            final AzureCredential finalCred = cred;
+            return azureConfigure().authenticate(new AzureTokenCredentials(env, null) {
```
Flanker32: could use `azureConfigure().authenticate(cred.getCredential())` here?

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/auth/AzureAuthHelper.java
```diff
@@ -160,10 +165,44 @@ protected Authenticated getAuthObj() {
...
+            AzureCredential cred = AzureLoginHelper.oauthLogin(env);
+            if (cred == null) {
+                // fallback to device login
+                cred = AzureLoginHelper.deviceLogin(env);
+            }
+            // if all tries fails
+            if (cred == null) {
+                return null;
+            }
+            final AzureCredential finalCred = cred;
```
Flanker32: why not use `cred` directly

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/auth/AzureCredential.java
```diff
@@ -0,0 +1,64 @@
...
+        if (authenticationResult == null) {
+            throw new NullPointerException("Parameter 'authenticationResult' cannot be null.");
+        }
+        this.env = env;
+        this.authenticationResult = authenticationResult;
+        this.refreshToken = authenticationResult.getRefreshToken();
+    }
+
+    public AzureTokenCredentials getCredential() {
+        return new AzureTokenCredentials(null, null) {
```
Flanker32: `new AzureTokenCredentials(env, "common")`, or we could make domain a constant

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/auth/oauth/AzureLoginHelper.java
```diff
@@ -0,0 +1,126 @@
...
+            System.out.println(String.format(LOGIN_MESSAGE_TEMPLATE,
+                    deviceCode.getVerificationUrl(), deviceCode.getUserCode()));
+            try {
+                // disable log4j of AuthenticationContext, otherwise the pending user authorization log
+                // will be print every second.
+                // see https://github.com/AzureAD/azure-activedirectory-library-for-java/issues/246
+                final Object logger = LoggerFactory.getLogger(AuthenticationContext.class);  
+                final Field levelField = logger.getClass().getSuperclass().getDeclaredField("currentLogLevel");
+                levelField.setAccessible(true);
+                levelField.set(logger, LocationAwareLogger.ERROR_INT + 1);
```
Flanker32: Shall we turn back the log level after device login?

andxu: Doesn't need since AuthenticationContext will be only used here

andxu:
This PR is on-hold since the final solution will be invoking `azure:login` goal.

PR #695, opened 2019-07-12
Replace windows line end to Linux line end.

andxu:


Flanker32:
Shall we add a `.gitattributes` to set the line endings?

PR #699, opened 2019-07-19
add unit test for auth helper

andxu:


azure-auth-helper/src/main/java/com/microsoft/azure/auth/AzureAuthHelper.java
```diff
@@ -52,6 +52,12 @@ public static AzureCredential deviceLogin(AzureEnvironment env) throws AzureLogi
      * @throws AzureLoginFailureException when there are some errors during refreshing.
      */
     public static AzureCredential refreshToken(AzureEnvironment env, String refreshToken) throws AzureLoginFailureException {
+        if (env == null) {
+            throw new NullPointerException("Parameter 'env' cannot be null.");
```
akaroml: This should be IllegalArgumentException instead of NPE which is too generic.

azure-auth-helper/src/main/java/com/microsoft/azure/auth/JsonUtils.java
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
+package com.microsoft.azure.auth;
```
akaroml: What is the purpose of adding another layer to wrap Gson?

PR #702, opened 2019-07-24
Fix NPE issue when no ApplicationInsights.xml found

wezhang:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/telemetry/AppInsightsProxy.java
```diff
@@ -75,7 +82,7 @@ private String readInstrumentationKeyFromConfiguration() {
             } else {
                 return StringUtils.EMPTY;
             }
-        } catch (IOException exception) {
+        } catch (IOException | NullPointerException exception) {
```
andxu: Don't catch NullPointerException to detect absent of file, use `if (inputStream == null) ` instead.

wezhang: Fix it.

PR #707, opened 2019-07-30
Rename some packages

andxu:
1. rename some package
2. add auth configuration
3. fix a bug of getAzureSecretFile(which will not read environment variables)
4. fix some javadoc issues


azure-auth-helper/src/test/java/com/microsoft/azure/auth/AzureAuthHelperTest.java
```diff
@@ -61,12 +61,8 @@ public void testRefreshTokenInvalidParameter() throws Exception {
 
     @Test
     public void tetGetAzureSecretFile() throws Exception {
-        File azureSecretFile = AzureAuthHelper.getAzureSecretFile();
-        assertEquals(Paths.get(System.getProperty(Constants.USER_HOME_KEY), ".azure", "azure-secret.json").toString(),
-                azureSecretFile.getAbsolutePath());
-        System.setProperty(Constants.AZURE_HOME_KEY, "test_dir");
-        azureSecretFile = AzureAuthHelper.getAzureSecretFile();
-        assertEquals(Paths.get("test_dir", "azure-secret.json").toFile().getAbsolutePath(),
+        final File azureSecretFile = AzureAuthHelper.getAzureSecretFile();
```
Flanker32: do we need to ensure `System.getenv(Constants.AZURE_CONFIG_DIR)` is empty or add correspond test case?

azure-auth-helper/src/test/java/com/microsoft/azure/auth/AzureAuthHelperTest.java
```diff
@@ -61,12 +61,8 @@ public void testRefreshTokenInvalidParameter() throws Exception {
...
+        assertEquals(Paths.get(System.getProperty(Constants.USER_HOME), ".azure", "azure-secret.json").toString(),
```
Flanker32: change to `Constants.AZURE_FOLDER` `Constants.AZURE_SECRET_FILE `

andxu: this is a test case, we are free to use actual constant values rather than Constants.AZURE_SECRET_FILE

azure-auth-helper/src/test/java/com/microsoft/azure/auth/AzureAuthHelperTest.java
andxu: java cannot set env itself, there is a workaround: https://stackoverflow.com/questions/318239/how-do-i-set-environment-variables-from-java

andxu: test case updated.

PR #700, opened 2019-07-22
add impl of oauth and device login

andxu:

PR #712, opened 2019-07-30
Add OAuth/device login and get token credentials logic

andxu:
This PR also contains the  changes in https://github.com/microsoft/azure-maven-plugins/pull/700


azure-auth-helper/src/main/java/com/microsoft/azure/auth/AzureLoginHelper.java
```diff
@@ -0,0 +1,239 @@
...
+        final LocalAuthServer server = new LocalAuthServer();
+        try {
+            server.start();
+            final URI redirectUri = server.getURI();
+            final String redirectUrl = redirectUri.toString();
+            final String code;
+            try {
+                final String authorizationUrl = authorizationUrl(env, redirectUrl);
+                Desktop.getDesktop().browse(new URL(authorizationUrl).toURI());
+                System.out.println(AUTH_WITH_OAUTH);
```
Flanker32: Can we use a logger here?

azure-auth-helper/src/main/java/com/microsoft/azure/auth/AzureServicePrincipleAuthHelper.java
```diff
@@ -0,0 +1,105 @@
...
+
+import java.io.File;
+import java.io.IOException;
+
+class AzureServicePrincipleAuthHelper {
+    private static final String AZURE_CLI_GET_SUBSCRIPTION_FAIL = "Failed to get default subscription of Azure CLI, please login Azure CLI first.";
+    private static final String AZURE_CLI_LOAD_TOKEN_FAIL = "Failed to load Azure CLI token file, " + "please login Azure CLI first.";
+
+    static AzureTokenCredentials getAzureServicePrincipleCredentials(AuthConfiguration config) throws InvalidConfigurationException, IOException {
+        if (StringUtils.isBlank(config.getClient())) {
```
Flanker32: What if user speicfy `serverId` and keep the secred in setting.xml?

PR #714, opened 2019-08-02
 add logic to read maven settings.

andxu:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/utils/TextUtils.java
```diff
@@ -25,4 +25,8 @@ public static String yellow(String message) {
     public static String green(String message) {
```
Flanker32: why not make the color as a parameter? And refactor other methods to `showWarningText` and `showConfirmText`

andxu: applyColorToText has done so

andxu: This method only append the color to the text, so 'showWarningText` is not proper.

PR #717, opened 2019-08-07
Remove port and change attributes in configuration to object

Flanker32:
Remove port to adapt new swagger change
Change attributes in configuration to object


azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/SpringConfiguration.java
```diff
@@ -10,20 +10,15 @@
 
 public class SpringConfiguration {
 
-    private int port;
-    private boolean isPublic;
+    private Boolean isPublic;
     private String subscriptionId;
     private String resourceGroup;
```
andxu: remove this?

Flanker32: We could remove it if we list cluster directly without resource group, this will be done in later pr.

PR #718, opened 2019-08-07
Update with latest swagger

Flanker32:
1. Use createOrUpdate only in create method
2. Use name as the identifer of deployment
3. Enable detail log within mvnDebug
4. Fix NPE issues

### DIFF
--- a/azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/AbstractSpringMojo.java
+++ b/azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/AbstractSpringMojo.java
@@ -10,8 +10,10 @@ import com.microsoft.azure.maven.spring.configuration.Deployment;
 import com.microsoft.azure.maven.spring.exception.SpringConfigurationException;
 import com.microsoft.azure.maven.spring.parser.SpringConfigurationParser;
 import com.microsoft.azure.maven.spring.parser.SpringConfigurationParserFactory;
+import com.microsoft.azure.maven.spring.spring.SpringServiceUtils;
 import com.microsoft.azure.maven.telemetry.AppInsightHelper;
 import com.microsoft.azure.maven.telemetry.MojoStatus;
+import com.microsoft.rest.LogLevel;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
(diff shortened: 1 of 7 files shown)


azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/utils/Utils.java
```diff
@@ -95,15 +97,13 @@ private static boolean isExecutableJar(File file) {
...
-        //            throw new MojoExecutionException(e.getMessage(), e);
-        //        }
-        //        return false;
+        try {
+            final CloudFile cloudFile = new CloudFile(new URI(sasUrl));
+            cloudFile.uploadFromFile(file.getPath());
+        } catch (Exception e) {
+            throw new MojoExecutionException(e.getMessage(), e);
+        }
+        return false;
```
andxu: why always returns false.

azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/AbstractSpringMojo.java
```diff
@@ -103,6 +105,11 @@ public void execute() throws MojoExecutionException, MojoFailureException {
     }
 
     protected void initExecution() {
+        // Init sdk log level
+        if (getLog().isDebugEnabled()) {
+            SpringServiceUtils.setLogLevel(LogLevel.BODY_AND_HEADERS);
```
andxu: will this introduce too many traces if users are running in debug mode -X?

azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/spring/SpringDeploymentClient.java
```diff
@@ -74,12 +78,27 @@ public DeploymentResourceInner createOrUpdateDeployment(
                 .withMemoryInGB(deploymentConfiguration.getMemoryInGB())
                 .withPersistentDisk(persistentDisk)
                 .withTemporaryDisk(temporaryDisk)
-                .withRuntimeVersion(RuntimeVersion.JAVA_8);
+                // Hard code Java 8 for now we support java 8 only
```
andxu: now we support java 8 only

azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/AbstractSpringMojo.java
Flanker32: Users could see the http requests and reponses within debug mode, it could be more convenient for debugging. (And i think there are already many traces in debug mode -X, it'l okay to add a bit more...

azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/utils/Utils.java
Flanker32: Fixed, this should be a void function

azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/spring/SpringDeploymentClient.java
Flanker32: Fixed

PR #727, opened 2019-08-12
Add validation code to auth configuration

andxu:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/validation/ConfigurationProblem.java
```diff
@@ -0,0 +1,48 @@
...
+        ERROR,
+        WARNING
+    }
+
+    private final String key;
+    private final String value;
+    private final String errorMessage;
+    private final Severity severity;
+
+
```
Flanker32: multi empty line

azure-auth-helper/src/main/java/com/microsoft/azure/auth/configuration/AuthConfiguration.java
```diff
@@ -89,4 +100,46 @@ public void setHttpProxyPort(String httpProxyPort) {
...
+            results.add(new ConfigurationProblem(null, null, "'httpProxyHost' and 'httpProxyPort' must both be set if you want to use proxy.", ERROR));
+        }
+
+        if (StringUtils.isNotBlank(httpProxyPort)) {
+            if (!StringUtils.isNumeric(httpProxyPort)) {
+                results.add(new ConfigurationProblem("httpProxyPort", httpProxyPort,
+                        String.format("Invalid integer number for httpProxyPort: '%s'.", httpProxyPort), ERROR));
+            } else if (NumberUtils.toInt(httpProxyPort) <= 0 || NumberUtils.toInt(httpProxyPort) > 65535) {
+                results.add(new ConfigurationProblem("httpProxyPort", httpProxyPort,
+                        String.format("Invalid range of httpProxyPort: '%s'.", httpProxyPort), ERROR));
```
Flanker32: We may add the valid range in the error message

azure-auth-helper/src/main/java/com/microsoft/azure/auth/configuration/AuthConfiguration.java
```diff
@@ -89,4 +100,46 @@ public void setHttpProxyPort(String httpProxyPort) {
         this.httpProxyPort = httpProxyPort;
     }
 
+    public List<ConfigurationProblem> validate() {
+        final List<ConfigurationProblem> results = new ArrayList<>();
+        final String errorMessagePostfix = StringUtils.isNotBlank(serverId) ? "server: " + serverId + " in settings.xml." : "<auth> configuration.";
```
Flanker32: return when `serverId!=null`? Or the prompt will be confusing if there are server id and wrong auth configuration

andxu: The <serverId> check should be done out of this class, when serverId is null, the caller must load a new AuthConfiguration instance and perform validation against the new AuthConfiguration instance, in AuthConfiguration  class, there is no maven related variables, so it is not possible to validate the property like 'serverId'

PR #728, opened 2019-08-12
rename a new field: jvmOptions and add the logic to save deployment to xml.__

andxu:


azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/utils/Utils.java
```diff
@@ -116,6 +107,37 @@ public static boolean isJarPackagingProject(MavenProject mavenProject) {
         return JAR.equalsIgnoreCase(mavenProject.getPackaging());
     }
 
+    public static void replaceDomWithKeyValue(Xpp3Dom node, String key, Object value) {
```
Flanker32: Move the Dom operations to DomHelper?

andxu: XmlUtils

PR #733, opened 2019-08-13
 rename javaVersion to runtimeVersion

andxu:


azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/AbstractSpringMojo.java
```diff
@@ -115,6 +135,30 @@ protected void initExecution() {
...
+                } catch (MavenDecryptException e) {
+                    throw new MojoFailureException(e.getMessage());
+                }
+            } else {
+                throw new MojoFailureException(String.format("Unable to get server('%s') from settings.xml.", auth.getServerId()));
+            }
+        }
+
+        final List<ConfigurationProblem> problems = auth.validate();
+        if (problems.stream().anyMatch(problem -> problem.getSeverity() == Severity.ERROR)) {
```
Flanker32: warning problems?

andxu: if there are only warning problems, the validation passes, but still should log the message, will cover this in newer PR.

PR #737, opened 2019-08-14
add prompter impl.

andxu:


azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/prompt/DefaultPrompter.java
```diff
@@ -0,0 +1,287 @@
...
+        return "(y/N)";
+    }
+
+    private static boolean isValidIntRangeInput(String text) {
+        return REGEX_COMMA_SEPARATED_INTEGER_RANGES.matcher(text).matches();
+    }
+
+    private static int[] parseIntRanges(String text, int minValue, int maxValue) {
+        final Matcher m = REGEX_NEXT_INTEGER_RANGE.matcher(text);
+        final List<Integer> values = new ArrayList<>();
```
Flanker32: Use a HashSet here, so that no need to remove duplicated items with stream later

azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/prompt/DefaultPrompter.java
```diff
@@ -0,0 +1,287 @@
...
+            final String input = reader.readLine();
+            if (StringUtils.isBlank(input)) {
+                if (hasDefaultValue || !isRequired) {
+                    return defaultValue;
+                }
+                System.out.print(emptyPromoteMessage);
+            } else {
+                try {
+                    return handleInput.apply(input);
+                } catch (UserInputException ex) {
```
Flanker32: Shall we use Exception to notify input error here? For it may impact the performance, shall we return null for validate error and return only if the result!=null?

azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/prompt/DefaultPrompter.java
```diff
@@ -0,0 +1,287 @@
...
+            }
+            throw new UserInputException(String.format(PROMPT_MESSAGE_STRING_TEMPLATE, input, name, finalHintMessage));
+        });
+    }
+
+    public Boolean promoteYesNo(Boolean defaultValue, String message, boolean isRequired) throws IOException {
+        final boolean hasDefaultValue = defaultValue != null;
+
+        if (!isRequired && hasDefaultValue) {
+            throw new IllegalArgumentException("There is no way to input empty value for a non-required field with default value.");
```
Flanker32: Seems we could remove the `isRequired` for we don't allow null value, and there will be no differences between required value and not required.

azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/prompt/DefaultPrompter.java
andxu: We need to return a value or an error, null cannot distinguish them, can use a tuple for two information.

azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/prompt/DefaultPrompter.java
andxu: done without hashset since it doesn't keep order.

PR #740, opened 2019-08-15
add unit test for auth-helper

andxu:


azure-auth-helper/src/test/java/com/microsoft/azure/auth/configuration/AuthConfigurationTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class AuthConfigurationTest {
+    @Test
+    public void testValidate() {
+        AuthConfiguration auth = new AuthConfiguration();
```
Flanker32: Add some comments here like what error will be throw here , as error count could be a bit confusing.

PR #748, opened 2019-08-19
Refactor SpringServiceUtils to SpringServiceClient

Flanker32:
* Refactor SpringServiceUtils to SpringServiceClient
* Change to use relativePath from resourceUploadDefinitionInner and set default deployment name to init


azure-spring-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/ConfigMojo.java
```diff
@@ -347,7 +345,7 @@ private void selectAppCluster() throws IOException, NoResourcesAvailableExceptio
 
     private void initializeCredentials() throws InvalidConfigurationException, IOException, NoResourcesAvailableException {
         // TODO: getAzureTokenCredentials will check auth for null, but maven will always map a default AuthConfiguration
-        final AzureTokenCredentials cred = AzureAuthHelper.getAzureTokenCredentials(this.auth);
+        final AzureTokenCredentials cred = azureTokenCredentials;
```
andxu: Doesn't need a special var cred

Issue #609, opened 2019-04-02
[Functions] Do not run func extensions install when using extension bundles


pragnagopa:
### Plugin name and version
Functions

Related archetype update: https://github.com/Microsoft/azure-maven-archetypes/issues/108

If an app is using bundles,  do not run `func extensions install` . This will eliminate the requirement for dotnet core / func cli at the app creation time.

cc: @soninaren @jeffhollan  @mattchenderson 

@jdneo / @Flanker32  - Please let us know if you have any questions.

Flanker32:
@pragnagopa  Sorry for reply late, so the maven plugin should detect whether the project using bundles and decide whether to execute `func extensions install`? Is there any schema here, or we just need to check `extensionBundle` exist in host.json?

pragnagopa:
> Is there any schema here, or we just need to check extensionBundle exist in host.json?

Yes you need to check host.json if the app is using extension bundle.

@soninaren - Can you please paste a sample host.json?

pragnagopa:
@soninaren will be updating CLI to take care of this scenario. Please hold off on this item until we have new CLI.

pragnagopa:
cc @amamounelsayed

@Flanker32  - Can you please verify if this issue is already addressed?

Flanker32:
@pragnagopa We are not aware of the status of extension bundles and function core tools and this issue is still in backlog. 

Besides, does this issue depends on https://github.com/microsoft/azure-maven-plugins/issues/658?

pragnagopa:
@Flanker32 - this is unrelated to issue #658 

if an app is using extension bundles, mvn plugin should not run function extensions install explicitly. @soninaren - Are we still waiting on CLI updates?

soninaren:
@pragnagopa The CLI updates required for this use are complete.

Here is the sample host.json file. You need to only check for the id property since the version can change.
```JSON
{
    "version": "2.0",
    "extensionBundle": {
        "id": "Microsoft.Azure.Functions.ExtensionBundle",
        "version": "[1.*, 2.0.0)"
    }
}
```

daniel-rocha:
Is there any way for me to test this change?

soninaren:
@daniel-rocha This is an optimization to make sure we don't call `func extensions install` where we don't need. To test this change, Once the plugin has been updated execute the `run`  command and check if the maven plugin is calling `func extensions install`

daniel-rocha:
[...]

PR #771, opened 2019-08-29
Support function bundles

Flanker32:
Skip func install when user specify function bundles, for #609

### DIFF
--- a/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
+++ b/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
@@ -314,7 +314,7 @@ public class PackageMojo extends AbstractFunctionMojo {
     protected boolean isInstallingExtensionNeeded(Set<BindingEnum> bindingTypes) {
         final JsonObject hostJson = readHostJson();
         final JsonObject extensionBundle = hostJson.getAsJsonObject(EXTENSION_BUNDLE);
-        if (extensionBundle != null &&
+        if (extensionBundle != null && extensionBundle.has("id") &&
                 StringUtils.equalsIgnoreCase(extensionBundle.get("id").getAsString(), EXTENSION_BUNDLE_ID)) {
             getLog().info(SKIP_INSTALL_EXTENSIONS_BUNDLE);
             return false;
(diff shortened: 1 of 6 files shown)


azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
```diff
@@ -287,11 +296,10 @@ protected FunctionCoreToolsHandler getFunctionCoreToolsHandler(final CommandHand
 
     protected void installExtension(final FunctionCoreToolsHandler handler,
                                     Set<BindingEnum> bindingEnums) throws Exception {
+        info(INSTALL_EXTENSIONS);
```
andxu: why move this stat up?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
```diff
@@ -304,8 +312,33 @@ protected void installExtension(final FunctionCoreToolsHandler handler,
     }
 
     protected boolean isInstallingExtensionNeeded(Set<BindingEnum> bindingTypes) {
-        return bindingTypes.stream().anyMatch(binding ->
+        final JsonObject hostJson = readHostJson();
+        final JsonObject extensionBundle = hostJson.getAsJsonObject(EXTENSION_BUNDLE);
+        if (extensionBundle != null &&
+                StringUtils.equalsIgnoreCase(extensionBundle.get("id").getAsString(), EXTENSION_BUNDLE_ID)) {
```
andxu: extensionBundle.get("id") will return null if id doesn't exist?

Flanker32: Yes, there may be NPE here, fixed

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
Flanker32: So that we will always show `Step 7 of 7: Installing function extensions if needed`
```
[INFO] Step 7 of 7: Installing function extensions if needed
[INFO] Extension bundle specified, skip install extension
```

pragnagopa:
@Flanker32 - Please tag @amamounelsayed  and @kulkarnisonia16 in future PRs for enhancements to functions maven plugin. Thanks!

PR #772, opened 2019-08-29
Fix the xml format issue

andxu:


andxu:
Close due to conflict

PR #801, opened 2019-09-09
add new project azure-maven-plugin-common to replace azure-maven-plug…

andxu:
…in-lib


azure-maven-plugin-common/src/main/java/com/microsoft/azure/maven/common/telemetry/AppInsightsProxy.java
```diff
@@ -0,0 +1,158 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class AppInsightsProxy implements TelemetryProxy {
```
Flanker32: This class is shared by Function/AppService Maven plugin which should not moved to common lib

andxu: This class is removed

PR #816, opened 2019-09-11
add test case for ConfigMojo

andxu:


azure-spring-cloud-maven-plugin/src/test/java/com/microsoft/azure/maven/spring/prompt/PromptWrapperTest.java
```diff
@@ -193,6 +201,54 @@ public void testBadTemplate() throws Exception {
...
+
+        try {
+            wrapper.handle("testId1", true, "foo");
+            fail("Should throw InvalidConfigurationException");
+        } catch (InvalidConfigurationException ex) {
+            // expected
+        }
+
+        try {
+            wrapper.handle("testId1", true, "foo");
```
Flanker32: Duplicate with the front one?

azure-spring-cloud-maven-plugin/src/test/resources/pom-1.xml
```diff
@@ -26,7 +26,7 @@
 		<plugins>
 			<plugin>
 				<groupId>com.microsoft.azure</groupId>
-				<artifactId>azure-spring-maven-plugin</artifactId>
+				<artifactId>azure-spring-cloud-maven-plugin</artifactId>
 				<version>0.0.1-SNAPSHOT</version>
```
Flanker32: Shall we add a parent pom or something else to get the correspond spring maven plugin version? Or we need to update every file when bump version

andxu: It is not good to use variable for artifactId

azure-spring-cloud-maven-plugin/src/test/java/com/microsoft/azure/maven/spring/prompt/PromptWrapperTest.java
andxu: Will delete it

PR #820, opened 2019-09-12
Fix Boolean NPE

Flanker32:
Fix NPE of Boolean when user didn't specify correspond value in pom.xml

### DIFF
--- a/azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/configuration/SpringConfiguration.java
+++ b/azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/configuration/SpringConfiguration.java
@@ -6,6 +6,8 @@
 
 package com.microsoft.azure.maven.spring.configuration;
 
+import org.apache.commons.lang3.BooleanUtils;
+
 public class SpringConfiguration {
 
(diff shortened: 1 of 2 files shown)


azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/configuration/SpringConfiguration.java
```diff
@@ -17,7 +19,7 @@
     private Deployment deployment;
 
     public Boolean isPublic() {
-        return isPublic;
+        return BooleanUtils.isTrue(isPublic);
```
andxu: Can we simple convert isPublic from Boolean to boolean

Flanker32:
> Please also fix dogfood boolean variable

will done in later pr

azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/configuration/SpringConfiguration.java
Flanker32: Leave it Boolean for it may break validation in config mojo, if there is no break, will convert them to boolean in later pr.

Issue #823, opened 2019-09-19
Running azure-spring-cloud:config can't authenticate with Device Login in linux consol


ZhengJun003:
### Plugin name and version
```
azure spring cloud maven plugin
0.1.0-SNAPSHOT
```
### Plugin configuration in your `pom.xml`
```
<plugin>
  <groupId>com.microsoft.azure</groupId>
  <artifactId>azure-spring-cloud-maven-plugin</artifactId>
  <version>0.1.0-SNAPSHOT</version>
</plugin>
```
### Expected behavior
![图片](https://user-images.githubusercontent.com/41051257/65223793-82a15f80-daf4-11e9-9330-8c1a163a8c3e.png)

### Actual behavior
![图片](https://user-images.githubusercontent.com/41051257/65223285-7072f180-daf3-11e9-8e28-3c9e67a3d04a.png)

ZhengJun003:
Fixed
Test in develop branch

PR #826, opened 2019-09-23
Fix issue of cannot device login 

andxu:


andxu:
https://github.com/microsoft/azure-maven-plugins/issues/823

PR #828, opened 2019-09-23
Rename default deployment name and set persistent storage path

Flanker32:
- Rename default deployment name to 'default'
- Set default mount path when enable persistent storage

### DIFF
--- a/azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/spring/SpringAppClient.java
+++ b/azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/spring/SpringAppClient.java
@@ -23,8 +23,9 @@ import java.util.stream.Collectors;
 
 public class SpringAppClient extends AbstractSpringClient {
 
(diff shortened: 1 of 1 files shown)


azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/spring/SpringAppClient.java
```diff
@@ -23,8 +23,9 @@
 
 public class SpringAppClient extends AbstractSpringClient {
 
-    protected static final String DEFAULT_DEPLOYMENT_NAME = "init";
     protected static final int DEFAULT_PERSISTENT_DISK_SIZE = 50;
+    protected static final String DEFAULT_DEPLOYMENT_NAME = "default";
```
andxu: move DEFAULT-XXX to class PersistentDisk

Flanker32: 'PersistentDisk' is the class in auto generated SDK, shall we change it?

PR #829, opened 2019-09-23
Record isPomFileModified

andxu:


azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/ConfigMojo.java
```diff
@@ -243,6 +245,7 @@ private void confirmAndSave() throws IOException {
     }
 
     private Integer saveConfigurationToPom() {
+        telemetries.put(TELEMETRY_KEY_POM_FILE_MODIFIED, String.valueOf(true));
```
Flanker32: Should we also record user discard the changes?

andxu: null indicates false.

PR #832, opened 2019-09-24
using maven 3.3.3 to keep consistency with other maven plugins.

andxu:


Flanker32:
Shall we bump the [annotation version](https://github.com/microsoft/azure-maven-plugins/blob/88ac4c63ae16aee44c04d697ea60a7d554fde687/azure-spring-cloud-maven-plugin/pom.xml#L141) and [maven-plugin-plugin](https://github.com/microsoft/azure-maven-plugins/blob/88ac4c63ae16aee44c04d697ea60a7d554fde687/azure-spring-cloud-maven-plugin/pom.xml#L200) version as well?

PR #834, opened 2019-09-25
Fix telemetry may not be sent within multi-project

Flanker32:
Fix telemetry may not be sent within multi-project, by calling getHttpClient instead of close.


azure-maven-plugin/src/main/java/com/microsoft/azure/plugin/login/AbstractAzureMojo.java
```diff
@@ -49,8 +49,6 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         } catch (Exception e) {
             handleException(e);
             throw e;
-        } finally {
-            AppInsightHelper.INSTANCE.close();
```
andxu: why remove this?

PR #837, opened 2019-09-26
Fix az login with sp and cert

andxu:


azure-auth-helper/src/main/java/com/microsoft/azure/auth/AzureServicePrincipleAuthHelper.java
```diff
@@ -50,7 +50,7 @@ static AzureTokenCredentials getAzureServicePrincipleCredentials(AuthConfigurati
      *
      * @return Authenticated object if Azure CLI 2.0 is logged with Service Principal.
      * @throws InvalidConfigurationException where there are some configuration errors
-     * @throws IOException where there read some read error when reading the file
+     * @throws IOException                   where there read some read error when reading the file
```
Flanker32: `tab` issue?

andxu: it is not a tab, but some spaces to make exception description start at the same position.

PR #843, opened 2019-09-30
 Use auth-helper to auth webapp&functions.

andxu:
Use auth-helper to auth webapp&functions.


azure-spring-cloud-maven-common/src/test/resources/pom-1.xml
```diff
@@ -0,0 +1,43 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
Flanker32: Why we move the resources to `spring-cloud-maven-common` ?

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
```diff
@@ -250,8 +280,50 @@ protected void initTelemetry() {
...
+            } else {
+                throw new AzureAuthFailureException(String.format("Unable to get server('%s') from settings.xml.", auth.getServerId()));
+            }
+        }
+
+        final List<ConfigurationProblem> problems = auth.validate();
+        if (problems.stream().anyMatch(problem -> problem.getSeverity() == Severity.ERROR)) {
+            throw new AzureAuthFailureException(String.format("Unable to validate auth configuration due to the following errors: %s",
+                    problems.stream().map(ConfigurationProblem::getErrorMessage).collect(Collectors.joining("\n"))));
+        }
```
Flanker32: Shall we move the check of server id as well as the validate to ` MavenSettingHelper.getAuthConfigurationFromServer`

andxu: Should be done later

azure-spring-cloud-maven-common/src/test/resources/pom-1.xml
andxu: Done

PR #846, opened 2019-10-10
update AI version

andxu:


Flanker32:
Do we need to update the ai version of plugin lib/common as well?

PR #851, opened 2019-10-15
Fix mac devices with touch bar share same installation id

Flanker32:
Fix mac devices with touch bar share same installation id, refers https://github.com/microsoft/vscode/pull/77033


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/telemetry/GetHashMac.java
```diff
@@ -25,6 +25,7 @@
 import java.security.MessageDigest;
```
Flanker32: `GetHasMac` of  azure-maven-plugin-lib is a simple copy of correspond class in azure-maven-plugin-common, the merge of maven-plugin-lib and maven-plugin-common is on the plan

PR #865, opened 2019-10-21
Fix issues in renaming jar artifact

Flanker32:
- Don't rename artifact when its name equals app.jar
- Use `File.renameTo` instead of `FileUtils.rename()` as `FileUtils` will delete the origin file when original file name equalsIgnoreCase with `app.jar`


andxu:
Please create the zip file in temp folder during deploying and try to delete it

andxu (approved):

PR #894, opened 2019-11-20
Support run from zip and run from blob

Flanker32:
Support 2 deployment methods, run from zip and run from blob

#### Run from zip
Run from package with app setting `WEBSITE_RUN_FROM_PACKAGE = 1`, this is recommanded for windows functions. We used to add `WEBSITE_RUN_FROM_PACKAGE = 1` in archetype and use zip deploy by default, however, some linux function hosting plan (consumption and premium) didn't support this config, so create a new deployment method.

#### Run from blob
Run from package with app setting `WEBSITE_RUN_FROM_PACKAGE = BlobURL`, plugin will upload function artifact to function storage and update the app setting, this is recommanded for linux consumption and premium plan.

### Extra jobs:
1. Move some common operations from `MSDeployArtifactHandlerImpl` to `FunctionArtifactHelper`
1. Some utils method, like get blob sas url, method convert `LocalDateTime` to `Date`
1. Add dependency for powermockito to test static methods


Flanker32:
Close to split it into smaller one

PR #898, opened 2019-11-20
Set the FUNCTIONS_EXTENSION_VERSION property to ~2

jdubois:
This should be the current default value, and it should be a constant so we can update it easily for our users


Flanker32:
@jdubois Thanks for your efforts. Now we set the default value of `FUNCTIONS_EXTENSION_VERSION ` in function archetype, you may find it [here](https://github.com/microsoft/azure-maven-archetypes/blob/develop/azure-functions-archetype/src/main/resources/archetype-resources/pom.xml#L133). 
Besides, shall we hard code the `FUNCTIONS_EXTENSION_VERSION` in maven plugin? As the V3 runtime of function is comming and i'm not sure whether users could switch their function runtime, so i prefer to keep the default value in archetype

jdubois:
@Flanker32  yes I know about the archetype, but I think it is better to have this within the plugin:

- It's cleaner for people using the archetype (they will wonder why they have this undocumented option)
- It will work for all people not using the plugin (including me, but also our friends from Pivotal who are using the plugin for Spring Cloud Functions)
- It will work for everybody who are upgrading from a previous release (this is the problem we just have in #897 as the integration tests where probably generated with an earlier version)

Of course I don't know your roadmap here, and you know the internals much better than I do, so this is just a proposal, you can close this if you don't think it is useful.

Flanker32:
@jdubois Thanks for your comments, i agree that it's more cleaner and convenient to have a default value in maven plugin, so i think we could add `FUNCTIONS_EXTENSION_VERSION` to `~2` as the default value if user didn't set it. But there are multi function [versions](https://docs.microsoft.com/en-us/azure/azure-functions/functions-versions) are avaiable, so i think we should not replace users' setting if they has already set it.

jdubois:
@Flanker32 this is just setting up a default value, not replacing what the user configured. It's the same mechanism that was already in place for another parameter, in fact I refactored the code so both use the same method now.

Flanker32:
@jdubois Plugin will call `getAppSettings()` and set it as the app settings, so call `setDefaultAppSetting` will override customers' app setting configuration.  The previous code is to set `FUNCTIONS_WORKER_RUNTIME` to `java`, as maven plugin only support java functions, so it will override other values directly, which differs from `FUNCTIONS_EXTENSION_VERSION`.
Could you please do a bit refactor like pass a boolean to indicate whether plugin should override customer setting so that both runtime and function version setting could share same code with different behavior.

jdubois:
Thanks @Flanker32 I'll have a look and refactor!

jdubois:
@Flanker32 thanks again! I have fixed the code, gave better names to the methods to avoid confusion, and added tests so we are sure this will be good.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
```diff
@@ -80,15 +84,34 @@
...
+        return settings;
+    }
+
+    private void forceDefaultAppSetting(Map result, String settingName, String settingIsEmptyMessage,
+                                        String settingValue, String changeSettingMessage) {
+
+        final String setting = (String) result.get(settingName);
+        if (StringUtils.isEmpty(setting)) {
+            info(settingIsEmptyMessage);
+        } else if (!settingValue.equals(setting)) {
```
andxu: settting.equals(settingVluae)

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
```diff
@@ -80,15 +84,34 @@
...
-            warning(String.format(CHANGE_FUNCTIONS_WORKER_RUNTIME, workerRuntime));
+        final Map settings = super.getAppSettings();
+        forceDefaultAppSetting(settings, FUNCTIONS_WORKER_RUNTIME_NAME, SET_FUNCTIONS_WORKER_RUNTIME,
+                FUNCTIONS_WORKER_RUNTIME_VALUE, CHANGE_FUNCTIONS_WORKER_RUNTIME);
+        setDefaultAppSetting(settings, FUNCTIONS_EXTENSION_VERSION_NAME, SET_FUNCTIONS_EXTENSION_VERSION,
+                FUNCTIONS_EXTENSION_VERSION_VALUE);
+        return settings;
+    }
+
+    private void forceDefaultAppSetting(Map result, String settingName, String settingIsEmptyMessage,
```
andxu: overrideDefaultAppSetting

jdubois:
Thanks @andxu  for the code review, I have fixed both issues

PR #899, opened 2019-11-22
Refactor web app runtime handler to share with function

Flanker32:
Refactor web app runtime handler to share with function
#### Description
`RuntimeHandler` is the Abstraction of creating app service, including create resource, update resource and update app service plan, webapp provide multi runtime handler implement to adapt multi runtime(win/linux/docker). As function is going to support linux/docker, move this part to plugin lib to avoid duplicate.
#### Details
1. Move `RuntimeHandler` from webapp to maven plugin lib, and make it generic to adapt WebApp and Function
2. Move `BaseRuntimeHandler` from webapp to maven plugin lib, will extract common part from `WebAppRuntimeHandler` in later PR


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/runtime/WebAppRuntimeHandler.java
```diff
@@ -15,14 +15,13 @@
...
 import com.microsoft.azure.management.resources.fluentcore.arm.Region;
-import com.microsoft.azure.maven.webapp.handlers.RuntimeHandler;
 import com.microsoft.azure.maven.webapp.utils.WebAppUtils;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.logging.Log;
 import org.apache.maven.settings.Settings;
 import org.codehaus.plexus.util.StringUtils;
 
-public abstract class BaseRuntimeHandler implements RuntimeHandler {
+public abstract class WebAppRuntimeHandler extends com.microsoft.azure.maven.handlers.runtime.BaseRuntimeHandler<WebApp> {
```
akaroml: why not import com.microsoft.azure.maven.handlers.runtime.BaseRuntimeHandler to avoid long line of code?

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/handlers/runtime/BaseRuntimeHandler.java
```diff
@@ -0,0 +1,28 @@
...
+
+package com.microsoft.azure.maven.handlers.runtime;
+
+import com.microsoft.azure.management.appservice.AppServicePlan;
+import com.microsoft.azure.management.appservice.WebAppBase;
+import com.microsoft.azure.maven.handlers.RuntimeHandler;
+
+public class BaseRuntimeHandler<T extends WebAppBase> implements RuntimeHandler<T> {
+    @Override
+    public WebAppBase.DefinitionStages.WithCreate defineAppWithRuntime() throws Exception {
```
akaroml: What does this class do? I was about to suggest making this an interface since its implementation is almost empty. But you already have RuntimeHandler interface. What is the point to have a man in the middle which does nothing?

Flanker32: I'll move the common part from webapp BaseRuntimeHandler to this class, including the common attributes and constructor, but it includes much code and test cases changes, will do it in next pr.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/runtime/WebAppRuntimeHandler.java
Flanker32: My mistake, this class used to called BaseRuntimeHandler and i didn't change the extends class after refactor it, will fix it in later commit.

PR #906, opened 2019-11-26
Add windows/linux runtime handler for function

Flanker32:
Add windows/linux runtime handler for function
  - Update existing function service plan will be added in later pr.


azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/handlers/runtime/WindowsFunctionRuntimeHandler.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public WebAppBase.Update updateAppRuntime(FunctionApp app) throws Exception {
+        return app.update();
+    }
+
+    @Override
+    public AppServicePlan updateAppServicePlan(FunctionApp app) throws Exception {
+
+        return null;
```
andxu: why return null?

Flanker32: It's a todo item, added comment in later commit.

Issue #908, opened 2019-11-26
azure-webapp-maven-plugin scales App Service plan (down) on deployment


viters:
### Plugin name and version
azure-webapp-maven-plugin 1.8.0

### Plugin configuration in your `pom.xml`
```xml
<plugin>
  <groupId>com.microsoft.azure</groupId>
  <artifactId>azure-webapp-maven-plugin</artifactId>
  <!-- check Maven Central for the latest version -->
  <version>${azure-webapp-maven-plugin.version}</version>
  <configuration>
    <schemaVersion>v2</schemaVersion>
    <resourceGroup>${azure.resource.group}</resourceGroup>
    <appName>${azure.app.name}</appName>
    <runtime>
      <os>linux</os>
      <javaVersion>jre8</javaVersion>
      <webContainer>tomcat 8.5</webContainer>
    </runtime>
    <deployment>
      <resources>
        <resource>
          <directory>${project.basedir}/target</directory>
          <includes>
            <include>*.war</include>
          </includes>
        </resource>
      </resources>
    </deployment>
  </configuration>
</plugin>
```

### Expected behavior
I have an existing Web App with existing App Service Plan. Plugin is configured to deploy to this specific app. This Web App and App Service Plan is configured. App Service Plan is P2V2. **Deployment with plugin should not change existing plans until explicitly configured**.

### Actual behavior
After deployment my existing App Service Plan is scaled down to P1V2. This process makes my other apps crash because of insufficient RAM.
**It bothers me that I have to configure each project and make sure all projects have correct pricingTier. Moreover, when I want to scale my plan I also have to update codebase of all projects using this plan.**

### Steps to reproduce the problem
Create an App Service Plan with P2V2. Create a Web App connected with this plan. Deploy to this plan with maven job with provided configuration. Check App Service Plan scaling.

Flanker32:
@viters Thanks for your report and sorry for the inconvenience, we will fix this issue soon.

Flanker32:
Fixed in #911

PR #910, opened 2019-11-27
#908 Skip scale of non-changing, existing App Service plan

viters:
As I have described in issue #908,

In my opinion an existing, non-changing App Service plan should not be scaled during azure-webapp-maven-plugin deployment process. It is confusing and makes plan scales hard to maintain.

I think the `<pricingTier>` should be considered only when App Service plan is created or switched, in other words in `If app's service plan differs from pom, change and update it` block.

Example situations:
1. I have couple of apps on one App Service plan. I have noticed enlarged traffic and scaled my plan up. Now I have to change `pom.xml` of every project, so the plan won't be scaled down after deployment of any of them.

2. I have many development apps on one plan, let's say 20 of them. I want to add additional ones, but it will require me to scale up plan. I need to update `pom.xml` of every app, so the every-night deployment will not scale plan down resulting in some of apps going down from insufficient resources.


Flanker32:
@viters Thanks for your contribution,in my opinion i prefer to keep scale pricingTier but only do this job only if pricingTier is specified in configuration. 
It's okay to skip pricingTier in configuration when deploying to existing app service plan, and pricing tier will not be updated while updating existing app service in this case. Besides, you still could scala the pricingTier by maven plugin if needed. And here is the link for [pr](https://github.com/microsoft/azure-maven-plugins/pull/911).

viters:
> @viters Thanks for your contribution,in my opinion i prefer to keep scale pricingTier but only do this job only if pricingTier is specified in configuration.
> It's okay to skip pricingTier in configuration when deploying to existing app service plan, and pricing tier will not be updated while updating existing app service in this case. Besides, you still could scala the pricingTier by maven plugin if needed. And here is the link for [pr](https://github.com/microsoft/azure-maven-plugins/pull/911).

I think you have got a point. I will close this PR then and vote for yours. Thank you.

PR #911, opened 2019-11-29
Support update app service plan in function

Flanker32:
1. Refactor BaseRuntimeHandler, support update app service plan in function maven plugin.
2. Update pricingTier only if user specify it in configuration, for #908

### DIFF
--- a/azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/handlers/runtime/BaseRuntimeHandler.java
+++ b/azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/handlers/runtime/BaseRuntimeHandler.java
@@ -20,7 +20,7 @@ import org.codehaus.plexus.util.StringUtils;
 
 public abstract class BaseRuntimeHandler<T extends WebAppBase> implements RuntimeHandler<T> {
 
-    private static final String TARGET_APP_SERVICE_PLAN_DO_NOT_EXIST = "Target app service plan %s was not founded in " +
+    private static final String TARGET_APP_SERVICE_PLAN_DO_NOT_EXIST = "Target app service plan %s cannot be found in " +
             "resource group %s, please check the configuration";
 
     protected String appName;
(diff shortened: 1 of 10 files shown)


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/handlers/runtime/BaseRuntimeHandler.java
```diff
@@ -12,12 +12,17 @@
...
 import com.microsoft.azure.maven.handlers.RuntimeHandler;
+import com.microsoft.azure.maven.utils.AppServiceUtils;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.logging.Log;
 import org.apache.maven.settings.Settings;
+import org.codehaus.plexus.util.StringUtils;
 
 public abstract class BaseRuntimeHandler<T extends WebAppBase> implements RuntimeHandler<T> {
 
+    private static final String TARGET_APP_SERVICE_PLAN_DO_NOT_EXIST = "Target app service plan %s was not founded in " +
```
andxu: cannot be found

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/utils/AppServiceUtils.java
```diff
@@ -81,4 +86,27 @@ public static String convertPricingTierToString(final PricingTier pricingTier) {
...
+        return webApp.manager().appServicePlans().getById(webApp.appServicePlanId());
+    }
+
+    public static AppServicePlan updateAppServicePlan(final AppServicePlan appServicePlan,
+                                                      final PricingTier pricingTier,
+                                                      final Log log) throws MojoExecutionException {
+        if (appServicePlan == null) {
+            throw new MojoExecutionException(SERVICE_PLAN_NOT_FOUND);
+        }
+        log.info(String.format(UPDATE_APP_SERVICE_PLAN));
```
andxu: "Updating app service plan" is also proper for functions?

Flanker32: Function is also a kind of app service deployed in app service plan, so i think it could be used for function too.

andxu: use 'Updating service plan' instead?

Flanker32: Portal also uses `App Service plan`, so it should be fine.
image

PR #913, opened 2019-12-03
Extract common logic for docker runtime handler

Flanker32:
1. Move docker image type and correspond util methods to plugin lib
2. Extract common codes to `AbstractLinuxFunctionRuntimeHandler` for docker


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/utils/AppServiceUtils.java
```diff
@@ -109,4 +110,20 @@ public static AppServicePlan updateAppServicePlan(final AppServicePlan appServic
     public static boolean isEqualAppServicePlan(AppServicePlan first, AppServicePlan second) {
         return first == null ? second == null : second != null && StringUtils.equals(first.id(), second.id());
     }
+
+    public static DockerImageType getDockerImageType(final String imageName, final String serverId,
+                                                     final String registryUrl) {
+        if (org.apache.commons.lang3.StringUtils.isEmpty(imageName)) {
```
andxu: use only one of org.codehaus.plexus.util.StringUtils or org.apache.commons.lang3.StringUtils in this class

Flanker32: Good catch, will fix it in later pr.

PR #935, opened 2019-12-11
use StringUtils in common-lang3

andxu:


azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/handlers/CommandHandlerImpl.java
```diff
@@ -106,7 +106,7 @@ protected void handleExitValue(int exitValue,
     protected void showErrorIfAny(final InputStream inputStream) throws Exception {
         if (inputStream != null) {
             final String input = IOUtil.toString(inputStream);
-            this.logger.error(org.codehaus.plexus.util.StringUtils.strip(input, "\n"));
+            this.logger.error(org.apache.commons.lang3.StringUtils.strip(input, "\n"));
```
Flanker32: Could remove the package name here

azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/spring/SpringAppClient.java
```diff
@@ -37,11 +38,11 @@ public Builder withAppName(String appName) {
             return self();
         }
 
-        public SpringAppClient build() {
+		public SpringAppClient build() {
```
Flanker32: Please use spaces instead of tab

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/handlers/CommandHandlerImpl.java
andxu: Done

azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/spring/SpringAppClient.java
andxu: Done

PR #946, opened 2019-12-17
WIP code split for refactoring function code to common

andxu:


azure-tools-common/src/main/java/com/microsoft/azure/common/function/IFunctionContext.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.microsoft.azure.auth.configuration.AuthConfiguration;
+import com.microsoft.azure.common.docker.IDockerCredentialProvider;
+import com.microsoft.azure.common.project.IProject;
+import com.microsoft.azure.maven.function.configurations.RuntimeConfiguration;
+
+public interface IFunctionContext {
+	String getDeploymentStagingDirectoryPath();
+
+	String getProjectDir();
```
andxu: IFunctionContext  should remove this method.

azure-tools-common/src/main/java/com/microsoft/azure/common/function/IFunctionContext.java
```diff
@@ -0,0 +1,38 @@
...
+
+	String getAppServicePlanName();
+
+	Map getAppSettings();
+
+	AuthConfiguration getAuth();
+
+	String getDeploymentType();
+
+	IDockerCredentialProvider getDockerCredentialProvider();
```
Flanker32: `IFunctionContext` should be the abstraction of function project, and plays a role like pojo? So i think it should extends `IDockerCredentialProvider` provide userName and password instead of provide `IDockerCredentialProvider`

azure-tools-common/src/main/java/com/microsoft/azure/common/project/IProject.java
```diff
@@ -0,0 +1,21 @@
...
+
+	Path getJarArtifact();
+
+	Path getBuildDirectory();
+
+	Path getClassesOutputDirectory();
+
+	Collection<Path> getProjectDepencencies();
+
+	boolean isWarProject();
```
Flanker32: Change to `Enum getProjectType()`?

andxu: not convenient when using it(will use getProjectType() == Enum.xxx)

azure-tools-common/src/main/java/com/microsoft/azure/common/function/IFunctionContext.java
andxu: When os is not docker, this provider will be null.

PR #950, opened 2019-12-18
Refactor main logic from mojo to common

andxu:


azure-tools-common/src/main/java/com/microsoft/azure/common/logging/Log.java
```diff
@@ -17,7 +17,7 @@ public static void error(String message) {
     }
 
     public static void info(String message) {
-        logger.info(message);
+        System.out.println(message);
     }
```
andxu: Because in maven, the info level will be print by default, while in gradle , info will not be print by default, here is a walkaround to show the messages, will finally change to notifyUser later

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AddMojo.java
```diff
@@ -6,550 +6,100 @@
...
+		this.functionName = StringUtils.capitalize(functionName);
+	}
+
+	protected void setFunctionTemplate(String functionTemplate) {
+		this.functionTemplate = functionTemplate;
+	}
+
+	@Override
+	protected void doExecute() throws Exception {
+		AddFunctionHandler handler = new AddFunctionHandler();
```
Flanker32: Mojo should do the job of asking for parameters and handler should only accept the inputs and generate function file

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AddMojo.java
```diff
@@ -6,550 +6,100 @@
...
+		handler.setBasedir(this.basedir);
+		handler.setCompileSourceRoots(compileSourceRoots);
+		handler.setFunctionName(functionName);
+		handler.setFunctionPackageName(functionPackageName);
+		handler.setFunctionTemplate(functionTemplate);
```
Flanker32: template should be saved in common

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/ListMojo.java
```diff
@@ -19,32 +21,9 @@
...
-        info(TEMPLATES_END);
-
-        info(BINDINGS_START);
-        printToSystemOut(BINDINGS_FILE);
-        info(BINDINGS_END);
-
-        info(RESOURCES_START);
-        printToSystemOut(RESOURCES_FILE);
-        info(RESOURCES_END);
+    	new ListHandler().execute();
```
Flanker32: We could keep `ListMojo` as this is only used for VSCode Function Extension, which will not be used in other toolkits.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/MavenFunctionContext.java
```diff
@@ -0,0 +1,149 @@
...
+	public String getDeploymentType() {
+		return mojo.getDeploymentType();
+	}
+
+	public IDockerCredentialProvider getDockerCredentialProvider() {
+		return getProvider(IDockerCredentialProvider.class);
+	}
+
+	@Override
+	public void registerProvider(Class<? extends Object> clazz, Object provider) {
```
Flanker32: clazz->class?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AddMojo.java
andxu: done

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/ListMojo.java
andxu: It is better to be in common, since it is simple without any dependency of maven.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/MavenFunctionContext.java
andxu: class is keyword

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AddMojo.java
andxu: good question, handler should keep some interfaces with the properties it requires.

azure-tools-common/src/main/java/com/microsoft/azure/common/function/PackageHandler.java
```diff
@@ -69,20 +69,58 @@
...
 		this.ctx = ctx;
 	}
 
+
+	private boolean checkStageDirectoryDirty() throws AzureExecutionException {
+		final File stagingDirectory = new File(ctx.getDeploymentStagingDirectoryPath());
+		File jarFileInStaging = new File(stagingDirectory, this.ctx.getProject().getJarArtifact().getFileName().toString());
+		File jarFileInBuild = this.ctx.getProject().getJarArtifact().toFile();
+		if (!jarFileInBuild.exists()) {
+			throw new AzureExecutionException("Artifact file '" + jarFileInBuild + "' cannot be found, please run 'gradle jar' to build the project.");
```
Flanker32: Should not throw exception with gradle commands in common

azure-tools-common/src/main/java/com/microsoft/azure/common/function/PackageHandler.java
```diff
@@ -69,20 +69,58 @@
...
+
+		if (!stagingDirectory.exists()) {
+			Log.info("Stage directory not found, build it now...");
+			return true;
+		}
+		if (stagingDirectory.exists() && stagingDirectory.isFile()) {
+			throw new AzureExecutionException("File name conflict: please delete file '" + stagingDirectory.getAbsolutePath() + "', and try again.");
+		}
+		if (jarFileInStaging.exists() && jarFileInStaging.length() == jarFileInBuild.length()) {
+			Log.info(STAGE_DIR_FOUND + ctx.getDeploymentStagingDirectoryPath());
```
Flanker32: We should also check whether host.json, function.json exist and properly, and i think it's okay to repackage if user do execute package, or we could add a force build parameter

azure-tools-common/src/main/java/com/microsoft/azure/common/function/PackageHandler.java
```diff
@@ -69,20 +69,58 @@
...
+		if (jarFileInBuild.length() == 0) {
+			throw new AzureExecutionException("Artifact file '" + jarFileInBuild + "' is empty, please run 'gradle jar' to build the project.");
+		}
+
+		if (!stagingDirectory.exists()) {
+			Log.info("Stage directory not found, build it now...");
+			return true;
+		}
+		if (stagingDirectory.exists() && stagingDirectory.isFile()) {
+			throw new AzureExecutionException("File name conflict: please delete file '" + stagingDirectory.getAbsolutePath() + "', and try again.");
```
Flanker32: I think we should split "whether we could package" and "whether we need package" test

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/artifact/WarArtifactHandlerImpl.java
```diff
@@ -103,7 +104,7 @@ protected String getContextPath() {
 
     protected File getWarFile() {
         return StringUtils.isNotEmpty(warFile) ? new File(warFile) :
-            new File(Paths.get(buildDirectoryAbsolutePath, project.getBuild().getFinalName() + ".war").toString());
+            new File(project.getJarArtifact().toString());
```
Flanker32: why get jar artifact here?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/MavenFunctionContext.java
Flanker32: so may be outputClass or some more meaningful name?

PR #956, opened 2019-12-23
Use slf4j api to log message

andxu:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/common/logging/Log.java
```diff
@@ -6,23 +6,22 @@
...
 import java.io.StringWriter;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
-public final class Log {
-    public static final String LOGGER_NAME = "azure-tools";
 
-    private static final Logger logger = Logger.getLogger(LOGGER_NAME);
+public final class Log {
+    private static Logger logger = org.slf4j.LoggerFactory.getLogger(Log.class);
```
Flanker32: shall we remove the full name here?

andxu: any name is proper

PR #959, opened 2019-12-24
update versions in pom.xml

andxu:


andxu:
The fix is at https://github.com/Azure/autorest-clientruntime-for-java/commit/d5b30ff98995ca76ce3e8754e5d6bcfbf868b34e#diff-02989d3b8308ec4591c357a4b94fc537

PR #962, opened 2019-12-24
move code of copying resources from handler to mojo

andxu:
2. use MavenDockerCredentialProvider to handle serverId


andxu:
Close because wrong target branch

PR #963, opened 2019-12-24
Refactor the logic of copying resources from artifact handler

andxu:
2. use MavenDockerCredentialProvider to handle serverId

### DIFF
--- a/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/DeployMojo.java
+++ b/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/DeployMojo.java
@@ -146,16 +146,7 @@ public class DeployMojo extends AbstractWebAppMojo {
             final ArtifactHandler artifactHandler = getFactory().getArtifactHandler(this);
             final boolean isV1Schema = SchemaVersion.fromString(this.getSchemaVersion()) == SchemaVersion.V1;
             if (isV1Schema) {
-                if (resources == null || resources.isEmpty()) {
-                    if (!(artifactHandler instanceof NONEArtifactHandlerImpl ||
-                            artifactHandler instanceof JarArtifactHandlerImpl ||
-                            artifactHandler instanceof WarArtifactHandlerImpl
-                            )) {
-                        throw new AzureExecutionException(NO_RESOURCES_CONFIG);
-                    }
-                } else {
-                    copyArtifactsToStagingDirectory(resources);
-                }
+                handleV1Resources(artifactHandler);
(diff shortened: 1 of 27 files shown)


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/DeployMojo.java
```diff
@@ -121,16 +143,94 @@ protected void deployArtifacts() throws AzureAuthFailureException, InterruptedEx
...
             }
-            getFactory().getArtifactHandler(this).publish(target);
+            final ArtifactHandler artifactHandler = getFactory().getArtifactHandler(this);
+            final boolean isV1Schema = SchemaVersion.fromString(this.getSchemaVersion()) == SchemaVersion.V1;
+            if (isV1Schema) {
+                handleV1Resources(artifactHandler);
+            } else {
+                final List<Resource> v2Resources = this.deployment == null ? null : this.deployment.getResources();
+
+                if (v2Resources == null || v2Resources.isEmpty()) {
```
Flanker32: why don't we check the resources in handle Function like V1?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/utils/Utils.java
```diff
@@ -8,20 +8,29 @@
...
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
 public class Utils {
 
     private static final String CREATE_TEMP_FILE_FAIL = "Failed to create temp file %s.%s";
 
+    private static final String NO_RESOURCES_CONFIG = "<resources> is empty. " + "Please make sure it is configured in pom.xml.";
```
Flanker32: Remove the +

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/DeployMojo.java
andxu: Because if (v2Resources == null || v2Resources.isEmpty()) the method deployArtifacts will return, in method handleV2Resources, there is no way to return for deployArtifacts

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/utils/Utils.java
andxu: Done

PR #970, opened 2019-12-31
Enhance auth library

Flanker32:
- prompt user auth method, auth file location and subscription name
- check subscription while get azure client


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/auth/AzureClientFactory.java
```diff
@@ -7,51 +7,82 @@
...
 public class AzureClientFactory {
+
+    private static final String SUBSCRIPTION_TEMPLATE = "Subscription : %s(%s)";
+    public static final String SUBSCRIPTION_NOT_FOUND = "Subscription %s was not found in current account";
+    public static final String NO_AVAILABLE_SUBSCRIPTION = "No available subscription found in current account";
+
     public static Azure getAzureClient(AuthConfiguration auth, String subscriptionId) throws InvalidConfigurationException, IOException,
         AzureLoginFailureException, InterruptedException, ExecutionException {
-        AzureTokenCredentials azureTokenCredentials = AzureAuthHelper.getAzureTokenCredentials(auth);
+        AzureTokenCredentialsDecorator azureTokenCredentials = AzureAuthHelper.getAzureTokenCredentials(auth);
```
andxu: azureTokenWrapper

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/auth/AzureClientFactory.java
```diff
@@ -7,51 +7,82 @@
...
+            checkSubscription(azure, subscriptionId);
+            final Subscription subscription = azure.getCurrentSubscription();
+            Log.info(String.format(SUBSCRIPTION_TEMPLATE, subscription.displayName(), subscription.subscriptionId()));
+            return azure;
         }
 
         return null;
     }
+
+    private static void checkSubscription(Azure azure, String subscriptionId) throws AzureLoginFailureException {
```
andxu: shall we also support subscription name?

andxu: if subscriptionId is empty , should return the first one if there are only one

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/auth/AzureClientFactory.java
```diff
@@ -7,51 +7,82 @@
...
             if (StringUtils.isEmpty(subscriptionId) && AzureAuthHelperLegacy.isInCloudShell()) {
                 subscriptionId = AzureAuthHelperLegacy.getSubscriptionOfCloudShell();
             }
-            return StringUtils.isEmpty(subscriptionId) ?
+            final Azure azure = StringUtils.isEmpty(subscriptionId) ?
                     authenticated.withDefaultSubscription() :
-                        authenticated.withSubscription(subscriptionId);
-
+                    authenticated.withSubscription(subscriptionId);
+            checkSubscription(azure, subscriptionId);
```
andxu: if subscriptionId is empty and there is default subscription id from "azure cli" etc, should use this default id

azure-auth-helper/pom.xml
```diff
@@ -8,7 +8,7 @@
         <relativePath>../azure-maven-plugins-pom</relativePath>
     </parent>
     <artifactId>azure-auth-helper</artifactId>
-    <version>0.2.0-SNAPSHOT</version>
+    <version>0.3.0-SNAPSHOT</version>
```
andxu: we have not released 0.2.0, keep 0.2.0-SNAPSHOT?

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/auth/AzureClientFactory.java
Flanker32: Java SDK only support subscription id, and we could only auth with azure first and then get the subscription name.

Flanker32: Java SDK will use the first subscription id if no default subscription

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/auth/AzureClientFactory.java
Flanker32: Will check default subscription and show warning if no subscription(default or pom) specified

azure-auth-helper/pom.xml
Flanker32: Fixed

akaroml:
Can we do one thing in one PR?
"Enhance auth library" doesn't provide clear info of what this PR is doing exactly.

PR #980, opened 2020-01-06
Show valid auth type in lower cases

Flanker32:
Show valid auth type in lower cases


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
```diff
@@ -291,7 +291,10 @@ protected AuthType getAuthTypeEnum() {
                 .filter(authTypeEnum -> StringUtils.equalsAnyIgnoreCase(authTypeEnum.name(), authType))
                 .findFirst().orElse(null);
         if (result == null) {
-            Log.warn(String.format(INVALID_AUTH_TYPE, authType, StringUtils.join(AuthType.values(), ", ")));
```
andxu: Please use simpler name for auth types other than strict enum values.

Flanker32: Fixed in #981

PR #986, opened 2020-01-07
Cache authTypeEnum in case prompt multi warnings

Flanker32:
Cache authTypeEnum in case prompt multi warnings when `getAuthTypeEnum()` called the second time.


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
```diff
@@ -299,20 +301,23 @@ protected Azure getAzureClientByAuthType() throws AzureAuthFailureException {
     }
 
     protected AuthType getAuthTypeEnum() {
```
andxu: wrap the logic of converting authType string to azure cred wrapper in a helper method.

Flanker32:
Fix in #987 , close this one.

Issue #991, opened 2020-01-09
Deploy slot with error: "Parameter publishingProfileOptions is required and cannot be null"


EricZhangzf:
### Plugin name and version
```
Maven plugin for azure webapps
1.9.0
```
### Plugin configuration in your `pom.xml`
```
<plugin> 
   <groupId>com.microsoft.azure</groupId> 
   <artifactId>azure-webapp-maven-plugin</artifactId> 
   <version>1.9.0</version> 
   <configuration> 
     <authType>auto</authType> 
     <subscriptionId>xxxxxxxxxxxxxxxxxxx</subscriptionId>
     <schemaVersion>V2</schemaVersion> 
     <resourceGroup>your-resource-group</resourceGroup> 
     <appName>web-20201</appName> 
     <pricingTier>P1v2</pricingTier> 
     <region>westeurope</region> 
     <runtime> 
        <os>linux</os> 
        <javaVersion>jre8</javaVersion> 
        <webContainer>WILDFLY 14</webContainer> 
     </runtime> 
     <deployment> 
        <resources> 
            <resource> 
               <directory>${project.basedir}/target</directory> 
               <includes> 
                   <include>*.war</include> 
               </includes> 
            </resource> 
         <resource>
             <directory>${project.basedir}</directory>
             <targetPath>/apache-tomcat-8.5.42/conf</targetPath>
             <includes>
                 <include>server.xml</include>
            </includes>
        </resource>
     </resources> 
     </deployment> 
     <deploymentSlot>
          <name>slot1</name>
          <configurationSource>PARENT</configurationSource>
     </deploymentSlot> 
     </configuration> 
 </plugin>
```
### Expected behavior
Deploy successfully and can access the URL
### Actual behavior
![image](https://user-images.githubusercontent.com/44559777/72053740-96bad800-3302-11ea-9480-9e07a644c9c6.png)

### Steps to reproduce the problem
1. `mvn azure-webapp:config` to generate the pom.xml
2. `mvn clean package`
3. `mvn azure-webapp:deploy ` to deploy the slot.

EricZhangzf:
Test pass

PR #992, opened 2020-01-10
Fix fail to get publishingProfile when deploy to deployment slot

Flanker32:
Fix error when deploy to Deploy slot "Parameter publishingProfileOptions is required and cannot be null", for issue #991


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/handlers/artifact/ArtifactHandlerImplV2.java
```diff
@@ -12,10 +12,12 @@
 import com.microsoft.azure.maven.appservice.OperatingSystemEnum;
 import com.microsoft.azure.maven.deploytarget.DeployTarget;
 import com.microsoft.azure.maven.handlers.artifact.ArtifactHandlerBase;
+import com.microsoft.azure.maven.utils.AppServiceUtils;
 import com.microsoft.azure.maven.webapp.configuration.RuntimeSetting;
 import com.microsoft.azure.maven.webapp.utils.FTPUtils;
 import com.microsoft.azure.maven.webapp.utils.Utils;
 
+import io.jsonwebtoken.lang.Collections;
```
andxu: bad util class

PR #994, opened 2020-01-10
Refactor: use common project in handler

andxu:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/common/project/IProject.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.Collection;
+
+public interface IProject {
+    Path getBaseDirectory();
+
+    Path getArtifactFile();
+
+    Path getBuildDirectory();
+
+    Path getClassesOutputDirectory();
```
Flanker32: `getClassesOutputDirectory` seems could be removed as common library will not do jobs depends on classes

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/common/project/IProject.java
```diff
@@ -0,0 +1,27 @@
...
+
+    Collection<Path> getProjectDepencencies();
+
+    boolean isWarProject();
```
Flanker32: `isWarProject()` `isJarProject()` could be implemented in abstract project or a util method

andxu: in future

PR #998, opened 2020-01-13
Move code from maven function plugin to azure-tools-common

andxu:


azure-tools-common/src/main/java/com/microsoft/azure/maven/function/handlers/FunctionCoreToolsHandlerImpl.java
```diff
@@ -28,25 +29,23 @@
     public static final String GET_LOCAL_VERSION_FAIL = "Failed to get Azure Functions Core Tools version locally";
     public static final Version LEAST_SUPPORTED_VERSION = Version.valueOf("2.0.1-beta.26");
 
-    private AbstractFunctionMojo mojo;
     private CommandHandler commandHandler;
 
-    public FunctionCoreToolsHandlerImpl(final AbstractFunctionMojo mojo, final CommandHandler commandHandler) {
-        this.mojo = mojo;
+    public FunctionCoreToolsHandlerImpl(final CommandHandler commandHandler) {
```
Flanker32: Do we need to add another constructor which add funcPath as parameter?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AddMojo.java
```diff
@@ -169,7 +169,7 @@ protected void doExecute() throws AzureExecutionException {
 
     //region Load templates
     protected BindingTemplate loadBindingTemplate(String type) {
-        try (final InputStream is = AddMojo.class.getResourceAsStream("/bindings.json")) {
+        try (final InputStream is = Constants.class.getResourceAsStream("/bindings.json")) {
```
Flanker32: I think it could be better to provide a Util method in common library to get the Function template

andxu: This PR is mainly for file move, let it be done in another PR

azure-tools-common/src/main/java/com/microsoft/azure/maven/function/handlers/FunctionCoreToolsHandlerImpl.java
andxu: This PR is mainly for file move, the changes in this java file is due to the bad reference to AbstractFunctionMojo

Issue #967, opened 2019-12-26
There is no artifact to deploy in staging directory when deploy.


EricZhangzf:
### Plugin name and version
azure-webapp-maven-plugin 1.9.0-SNAPSHOT
### Plugin configuration in your `pom.xml`
```
<build> 
    <plugins> 
      <plugin>
        <groupId>com.microsoft.azure</groupId>
        <artifactId>azure-webapp-maven-plugin</artifactId>
        <version>1.9.0-SNAPSHOT</version>
        <configuration>
          <schemaVersion>V2</schemaVersion>
          <resourceGroup>your-resource-group</resourceGroup>
          <appName>your-appname</appName>
          <pricingTier>P1v2</pricingTier>
          <region>eastus</region>
          <runtime>
            <os>linux</os>
            <javaVersion>jre8</javaVersion>
            <webContainer>tomcat 8.5</webContainer>
          </runtime>
          <deployment>
            <resources>
              <resource>
                <directory>${project.basedir}/target</directory>
                <includes>
                  <include>*.war</include>
                </includes>
              </resource>
            </resources>
          </deployment>
        </configuration>
      </plugin>
    </plugins> 
  </build> 
```
### Actual behavior
![image](https://user-images.githubusercontent.com/44559777/71461848-49ac1180-27ec-11ea-99bb-ba7e137c92d2.png)

### Steps to reproduce the problem
1.  mvn azure-webapp:config to config your pom.xml like above.
2.  mvn clean package
3.  mvn azure-webapp:deploy

EricZhangzf:
Test pass on 1.9.0

EricZhangzf:
Still repro on 1.10.0-SNAPSHOT three OS, So reopen it.

Flanker32:
Fixed in #1004

PR #1004, opened 2020-01-15
fix a bug when there are only external resources.

andxu:


andxu:
https://github.com/microsoft/azure-maven-plugins/issues/967

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/DeployMojo.java
```diff
@@ -153,19 +153,24 @@ protected void deployArtifacts(WebAppConfiguration webAppConfiguration)
...
+                }
             }
             artifactHandler.publish(target);
         } finally {
             util.afterDeployArtifacts();
         }
     }
 
-    private void handleV2Resources(final DeployTarget target, List<Resource> v2Resources) throws IOException, AzureExecutionException {
+    private List<Resource> handleV2Resources(final DeployTarget target, List<Resource> v2Resources) throws IOException, AzureExecutionException {
```
Flanker32: We may need to rename this method as a handle method should not return a List with specified values, or we could move `artifactHandler.publish(target);` to handle methods

PR #1018, opened 2020-02-11
Update to checkstyle 8.29

andxu:
since the code change at https://github.com/checkstyle/checkstyle/issues/7119 causing too many *custom.import.order.separated.internally* violations, I set the import check to info level

### DIFF
--- a/azure-auth-helper/pom.xml
+++ b/azure-auth-helper/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>com.microsoft.azure</groupId>
         <artifactId>azure-maven-plugins-pom</artifactId>
-        <version>1.0.2</version>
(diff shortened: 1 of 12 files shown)


azure-auth-helper/pom.xml
```diff
@@ -4,7 +4,7 @@
     <parent>
         <groupId>com.microsoft.azure</groupId>
         <artifactId>azure-maven-plugins-pom</artifactId>
-        <version>1.0.2</version>
+        <version>1.0.3</version>
```
Flanker32: 1.0.3-SNAPSOHT?

andxu: NO, this pom will not changed frequently

Issue #1008, opened 2020-01-16
Webapp/Functions will do useless during deployment update when the settings are not changed


andxu:

Flanker32:
Fixed in #1022

PR #1022, opened 2020-02-14
Update app service plan only when necessary

Flanker32:
Update app service plan only when necessary, for #1008


azure-tools-common/src/main/java/com/microsoft/azure/common/utils/AppServiceUtils.java
```diff
@@ -101,13 +101,13 @@ public static AppServicePlan updateAppServicePlan(final AppServicePlan appServic
         if (appServicePlan == null) {
             throw new AzureExecutionException(SERVICE_PLAN_NOT_FOUND);
         }
+        // Return when no need to update
+        if (pricingTier == null || appServicePlan.pricingTier().equals(pricingTier)) {
```
andxu: appServicePlan.pricingTier() will never be null?

PR #1038, opened 2020-03-17
Move more common codes to azure-tools-common

andxu:


azure-maven-plugin-common/pom.xml
```diff
@@ -14,7 +14,7 @@
     <packaging>jar</packaging>
     <name>Azure Maven Plugin Common Library</name>
     <description>Common
```
Flanker32: why we make a new line here

andxu: Removed

PR #1060, opened 2020-04-30
fix NPE exception when the 'runtimeVersion' is not set

jdubois:
In the Maven plugin, if the `<runtimeVersion>Java8</runtimeVersion>` is not set (which should be the majority of cases, as it's an optional parameter, and as only Java 8 is supported at the moment according to the doc - when in fact Java 11 also works), then you'll get an NPE when doing the regexp on the line below.
This prevents the NPE, and instead gives the default value, which should be the case when nothing is configured.

### DIFF
--- a/azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/spring/SpringDeploymentClient.java
+++ b/azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/spring/spring/SpringDeploymentClient.java
@@ -106,6 +106,9 @@ public class SpringDeploymentClient extends AbstractSpringClient {
     }
 
     private RuntimeVersion parseRuntimeVersion(String runtimeVersion) {
+        if (StringUtils.isAllEmpty(runtimeVersion)) {
+            return DEFAULT_RUNTIME_VERSION;
+        }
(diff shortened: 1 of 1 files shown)


jdubois:
BTW I said something wrong, only Java 8 is supported in the plugin... When Java 11 is supported on the platform - do you want another PR for this? There are several things I would like to improve here

Flanker32:
> said something wrong, only Java 8 is supported in the plugin... When Java 11 is supported on the platform - do you want another PR for this? There are several things I would like to

Hi jdubois, java 11 is supported in both service and maven plugin, is there any issue with java 11 runtime in spring maven plugin?

jdubois:
Oh sorry @Flanker32 I must have been on another branch, or distracted: it does support Java 8 and Java 11.
Now what I don't like (and that might have been my original issue in fact), is that it's not the same String as with the `az` command line: it's `Java8` or `java8`, and not `java_8`. Maybe we could accept all 3 options? Anyway, the error message should give the correct possible answers: here it just says it's wrong, and there is no way to know what the correct options are, unless you read the source code.

Flanker32:
@jdubois We are using `Java 8` and `Java 11` as the supported values to be consistent with azure portal, here is the screen shot.
image

I think its better to support both the three cases so that users will not be bothered by the wording, i'll create a pr for this and fix the error message, thanks a lot for your advices.

PR #1082, opened 2020-06-04
Add support for java 11 functions

Flanker32:
Add support for java 11 functions


andxu:
Add telemetry for java 8 or 11.

azure-tools-common/src/main/java/com/microsoft/azure/common/function/handlers/runtime/WindowsFunctionRuntimeHandler.java
```diff
@@ -59,11 +66,25 @@ protected WindowsFunctionRuntimeHandler(Builder builder) {
...
         }
+        appWithCreate.withJavaVersion(javaVersion).withWebContainer(null);
         return appWithCreate;
     }
 
     @Override
     public FunctionApp.Update updateAppRuntime(FunctionApp app) {
-        return app.update();
+        final JavaVersion serverJavaVersion = app.javaVersion();
+        final FunctionApp.Update update = app.update();
```
andxu, line 76: handle null serverJavaVersion

azure-tools-common/src/main/java/com/microsoft/azure/common/function/handlers/runtime/FunctionRuntimeHandler.java
```diff
@@ -72,4 +80,26 @@ protected void changeAppServicePlan(FunctionApp app, AppServicePlan appServicePl
     protected ResourceGroup getResourceGroup() {
         return azure.resourceGroups().getByName(resourceGroup);
     }
+
+    protected JavaVersion parseJavaVersion() {
```
andxu, line 84: protected to private

azure-tools-common/src/main/java/com/microsoft/azure/common/function/handlers/runtime/WindowsFunctionRuntimeHandler.java
```diff
@@ -59,11 +66,25 @@ protected WindowsFunctionRuntimeHandler(Builder builder) {
...
             }
         }
+        appWithCreate.withJavaVersion(javaVersion).withWebContainer(null);
         return appWithCreate;
     }
 
     @Override
     public FunctionApp.Update updateAppRuntime(FunctionApp app) {
-        return app.update();
+        final JavaVersion serverJavaVersion = app.javaVersion();
```
andxu, line 75: when local configuration has not configured java version and shall we update Java version at server side?  

```
if (javaVersion.equals(serverJavaVersion)) {
            Log.info(String.format(HOST_JAVA_VERSION, serverJavaVersion));
        } else if (JavaVersion.OFF.equals(serverJavaVersion)) {
            Log.info(HOST_JAVA_VERSION_OFF);
            update.withJavaVersion(javaVersion);
        }
```

Flanker32: Will update the server side java version only if user set javaVersion in `pom.xml`
``` java
else if (StringUtils.isNotEmpty(runtimeConfiguration.getJavaVersion())) {
            // Will update function host java version if user specify it in pom
            Log.info(String.format(HOST_JAVA_VERSION_UPDATE, serverJavaVersion, javaVersion));
            update.withJavaVersion(javaVersion);
        }
```

PR #1090, opened 2020-06-18
Adding EventGrid output binding

vrdmr:
Adding the new EventGrid output binding - Added here - https://github.com/Azure/azure-functions-java-library/pull/112.


vrdmr:
@amamounelsayed FYI

PR #1136, opened 2020-08-29
1135: Read the 'finalName' from plugin configuration to allow customisation by the user

patrickherrera:
Also improve logging to clarify what file is actually being looked for

Fixes: https://github.com/microsoft/azure-maven-plugins/issues/1135


andxu (approved):

andxu:
@patrickherrera, your code LGTM, thanks for contributing

dante-ricalde:
I have tested the plugin 1.10.0-SNAPSHOT and it's validating againts the jar name...I think this validation is only taking into account jar files...in my case I was using the version 1.6.0 to deploy a graal native image to azure functions and it was working ok...but now using the version 1.10.0-SNAPSHOT from branch develop is not working because it's validating against a jar file...but I don't have a jar file ...I have a native image like this for example: mvn-quarkus-azure-graal-function-example-1.0-SNAPSHOT-runner

PR #81, opened 2018-01-18
Make project friendly fro Eclipse users

dsyer:

PR #1140, opened 2020-09-09
[devops #1764044] unify dependency management for `azure-maven-plugins`

wangmingliang-ms:
Task 1769084: manage dependencies in root project's pom.


azure-tools-common/pom.xml
```diff
@@ -85,87 +58,74 @@
...
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>${jackson.core.version}</version>
         </dependency>
         <dependency>
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
-            <version>${commons.io.version}</version>
+        </dependency>
+        <dependency>
```
Flanker32, line 90: Is there any code changes to add a new dependency here?

azure-webapp-maven-plugin/pom.xml
```diff
@@ -317,30 +281,6 @@
                     </execution>
                 </executions>
             </plugin>
-            <plugin>
```
Flanker32, line 320: https://github.com/microsoft/azure-maven-plugins/pull/81, this is for better development experience in Eclipse, do we need to remove it?

azure-tools-common/pom.xml
wangmingliang-ms: no, but it's explicitly used in code. no sure why it's not explicitly declared.
```
com/microsoft/azure/common/function/handlers/runtime/DockerFunctionRuntimeHandler.java:18
```
```
com/microsoft/azure/common/utils/TextUtilsTest.java:8
```

azure-webapp-maven-plugin/pom.xml
wangmingliang-ms: I will restore it.

andxu (approved):

PR #1147, opened 2020-09-18
replace MavenPluginQueryerDefaultImpl with TextIO 

andxu:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/queryer/TextIOMavenPluginQueryer.java
```diff
@@ -0,0 +1,58 @@
...
+            Log.info(FOUND_VALID_VALUE);
+            return initValue;
+        }
+        prompt = StringUtils.isEmpty(prompt) ? getPromptString(attribute) : prompt;
+        return textIO.<String>newGenericInputReader(null)
+                .withNumberedPossibleValues(options).withDefaultValue(defaultValue).read(prompt);
+    }
+
+    @Override
+    public String assureInputFromUser(String attribute, String defaultValue, String regex,
```
Flanker32, line 38: It seems that we could remove `errorMessage` parameter, and the value is never specified in caller

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/queryer/TextIOMavenPluginQueryer.java
```diff
@@ -0,0 +1,58 @@
...
+            Log.info(FOUND_VALID_VALUE);
+            return initValue;
+        }
+
+        prompt = StringUtils.isEmpty(prompt) ? getPromptString(attribute) : prompt;
+        return textIO.newStringInputReader().withPattern(regex).withDefaultValue(defaultValue).read(prompt);
+    }
+
+    private String getPromptString(String attributeName) {
+        return String.format(PROMPT_STRING_WITHOUT_DEFAULTVALUE, attributeName);
```
Flanker32, line 51: did textIO will prompt default value by default?

andxu: no, we need to provide

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/queryer/TextIOMavenPluginQueryer.java
andxu: Then we need to refactor ConfigMojo, will be present in another PR

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/queryer/TextIOMavenPluginQueryer.java
andxu: NO

Flanker32 (approved):

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/queryer/TextIOMavenPluginQueryer.java
andxu: textIO will provide the fault value for string input, but for list options, it will only print a start(*). Will refactor text io for prompt message in later PR

PR #1154, opened 2020-09-21
support serializing subscription id and service plan

andxu:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
```diff
@@ -300,10 +300,13 @@ protected Azure getAzureClientByAuthType() throws AzureAuthFailureException {
                 environment = AzureAuthHelper.getAzureEnvironment(auth.getEnvironment());
             }
             final String environmentName = AzureAuthHelper.getAzureEnvironmentDisplayName(environment);
-            if (environment != AzureEnvironment.AZURE) {
+            if (environment != AzureEnvironment.AZURE && azure == null) {
```
shenqianjin, line 303: how about using AzureEnvironment.AZURE.equals(environment) ?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
```diff
@@ -136,7 +146,7 @@ protected boolean confirmConfiguration(WebAppConfiguration configuration) throws
             System.out.println("Slot name : " + slotSetting.getName());
             System.out.println("ConfigurationSource : " + slotSetting.getConfigurationSource());
         }
-        final String result = queryer.assureInputFromUser("confirm", "Y", BOOLEAN_REGEX, "Confirm (Y/N)? : ", null);
+        final String result = queryer.assureInputFromUser("confirm", "Y", BOOLEAN_REGEX, "Confirm (Y/N)", null);
```
shenqianjin, line 149: how about replacing the values of string (such as "Y", "N", etc) with constant variables?

azure-tools-common/src/main/java/com/microsoft/azure/common/Utils.java
```diff
@@ -84,4 +87,16 @@ public static String getSegment(String id, String segment) {
         }
         return null;
     }
+
+    public static boolean isPomPackagingProject(String packaging) {
```
Flanker32, line 91: create an enum for packaging and use fromString here as instead?

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/auth/AzureClientFactory.java
```diff
@@ -31,22 +31,17 @@
...
+        return StringUtils.isEmpty(subscriptionId) ? authenticated.withDefaultSubscription() :
                 authenticated.withSubscription(subscriptionId);
-        checkSubscription(azureClient, subscriptionId);
-        final Subscription subscription = azureClient.getCurrentSubscription();
-        Log.info(String.format(SUBSCRIPTION_TEMPLATE, subscription.displayName(), subscription.subscriptionId()));
-        return azureClient;
     }
 
-    private static void checkSubscription(Azure azure, String targetSubscription) throws AzureLoginFailureException {
+    public static void checkSubscription(Azure azure, String targetSubscription) throws AzureLoginFailureException {
```
Flanker32, line 44: If we separate `checkSubscription` from `getAzureClient`, do we need to add  `checkSubscription` back to `getAzureClient` caller

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/serializer/V2ConfigurationSerializer.java
```diff
@@ -23,9 +23,10 @@ public V2ConfigurationSerializer(WebAppConfiguration newConfigs, WebAppConfigura
     }
 
     @Override
-    public void saveToXML(Element configurationElement)
+    public void saveToXML(Element configurationElement, String subscriptionId)
         throws MojoFailureException {
         createOrUpdateAttribute("schemaVersion", "V2", oldConfigs.getSchemaVersion(), configurationElement);
+        createOrUpdateAttribute("subscriptionId", subscriptionId, null, configurationElement);
```
Flanker32, line 29: creating `<subscription>` node directly may override users' parameters, like `<subscriptionId>${subs}</subscriptionId>`, so I prepare to use `createOrUpdateAttribute` to compare with previous configuration and save only if the value changed

azure-tools-common/src/main/java/com/microsoft/azure/common/Utils.java
andxu: maven itself uses string for handing packing, it is not necessary to create a enum, since packing is not limited to jar war

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/auth/AzureClientFactory.java
andxu: I will roll this change back

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
```diff
@@ -60,12 +59,22 @@
...
 
+    private static final String PRICE_TIER_NOT_AVAIL = "The price tier \"P1\", \"P2\", \"P3\" are only available for Windows runtime, use \"%s\" instead.";
+
     private MavenPluginQueryer queryer;
     private WebAppPomHandler pomHandler;
     private static final String[] configTypes = {"Application", "Runtime", "DeploymentSlot"};
 
     @Override
     protected void doExecute() throws AzureExecutionException {
+        if (!(Utils.isPomPackagingProject(this.project.getPackaging()) ||
```
Flanker32, line 70: could we create a new static filed store all the supported packaging and validate by `StringUtils::equalsAny` ?

Flanker32 (approved):

PR #1158, opened 2020-09-23
adjust printCurrentSubscription to getAzureClient and add some utilit…

andxu:
…y functions to AppServiceUtils

### DIFF
--- a/azure-tools-common/src/main/java/com/microsoft/azure/common/utils/AppServiceUtils.java
+++ b/azure-tools-common/src/main/java/com/microsoft/azure/common/utils/AppServiceUtils.java
@@ -14,9 +14,7 @@ import com.microsoft.azure.management.appservice.AppServicePlan;
 import com.microsoft.azure.management.appservice.OperatingSystem;
 import com.microsoft.azure.management.appservice.PricingTier;
 import com.microsoft.azure.management.appservice.RuntimeStack;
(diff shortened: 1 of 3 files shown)


azure-tools-common/src/main/java/com/microsoft/azure/common/utils/AppServiceUtils.java
```diff
@@ -124,4 +128,33 @@ public static DockerImageType getDockerImageType(final String imageName, final b
...
+        }
+        return new RuntimeStack(segments[0], segments[1]);
+    }
+
+    public static boolean isDockerWebapp(WebApp webapp) {
+        final String linuxFxVersion = webapp.linuxFxVersion();
+        return StringUtils.containsIgnoreCase(linuxFxVersion, "DOCKER|");
+    }
+
+    private static String getSegment(String id, String segment) {
```
Flanker32, line 152: There has been a same method in utils https://github.com/microsoft/azure-maven-plugins/blob/develop/azure-tools-common/src/main/java/com/microsoft/azure/common/Utils.java#L79

andxu: Done

Flanker32 (approved):

PR #1163, opened 2020-09-24
Begin using git token with Rich Nav task

jepetty:
A recent update to our service requires a Git token with all indexing requests. With this change, the Rich Nav task will have access to the aforementioned token to make requests to our service.


andxu (approved):

andxu:
Thank you @jepetty and @xuachen.

PR #1164, opened 2020-09-24
Update RichNav action to acquire github token

xuachen:
Another recent change missed before which requires getting the github token to successfully index the repo. This will prevent the RichNav action from throwing 401 errors.


xuachen:
A duplicate of https://github.com/microsoft/azure-maven-plugins/pull/1163

PR #1172, opened 2020-10-14
support new style configuration for runtime and jboss

andxu:
1. add jboss support
2. add price selection during create wizard.
3. change to standard value of webContainer and javaVersion
Sample config:
```
 <runtime>
      <os>Linux</os>      
      <webContainer>Java SE</webContainer>
      <javaVersion>Java 11</javaVersion>
    </runtime>
```

Add jboss support

### DIFF
--- a/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
+++ b/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
@@ -433,10 +433,6 @@ public class ConfigMojo extends AbstractWebAppMojo {
         // For project which package is jar, use java se runtime
         if (isJarProject()) {
             final WebContainer webContainer = WebContainerUtils.getJavaSEWebContainer();
-            if (Objects.isNull(webContainer)) {
-                throw new MojoFailureException(
-                        String.format("Java SE environment is not supported in %s", javaVersion.toString()));
-            }
             return builder.javaVersion(javaVersion).webContainer(webContainer);
         }
         final String defaultWebContainer = configuration.getWebContainerOrDefault();
--- a/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
+++ b/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
@@ -341,27 +341,23 @@ public class ConfigMojo extends AbstractWebAppMojo {
         }
         final String javaVersion = queryer.assureInputFromUser("javaVersion", defaultJavaVersion,
                 JavaVersionUtils.getValidJavaVersions(), null);
-        // For project which package is jar, use java se runtime2
+        // For project which package is jar, use java se runtime
         if (isJarProject()) {
             return builder.runtimeStack(RuntimeStackUtils.getJavaSERuntimeStack(javaVersion));
         }
         final List<String> validRuntimeStacks = RuntimeStackUtils.getValidLinuxRuntimeStacksForJavaVersion(javaVersion);
-        String defaultLinuxRuntimeStack = StringUtils.capitalize(configuration.getLinuxRuntimeStackOrDefault());
+        String defaultLinuxRuntimeStack = configuration.getLinuxRuntimeStackOrDefault();
         // sometimes the combination of <javaVersion> and <webContainer> may not be valid, but
         // <webContainer> might be right
         if (Objects.isNull(configuration.getRuntimeStack()) && Objects.nonNull(this.getRuntime()) &&
                  StringUtils.isNotBlank(this.getRuntime().getWebContainerRaw())) {
-            final String validRuntStackFromExisting = findStringInCollectionIgnoreCase(validRuntimeStacks, this.getRuntime().getWebContainerRaw());
-            if (validRuntStackFromExisting == null) {
+            final String validRuntStackFromExistingConfiguration = findStringInCollectionIgnoreCase(validRuntimeStacks, this.getRuntime().getWebContainerRaw());
+            if (validRuntStackFromExistingConfiguration == null) {
                 Log.warn(String.format("'%s' is not supported in java version: %s",
                         this.getRuntime().getWebContainerRaw(), javaVersion));
             } else {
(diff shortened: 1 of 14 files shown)


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
```diff
@@ -142,6 +152,36 @@ protected void config(WebAppConfiguration configuration) throws MojoFailureExcep
...
+        if (Objects.nonNull(configuration) && Objects.nonNull(result.getOs()) && Objects.nonNull(this.getRuntime())) {
+            switch (result.getOs()) {
+                case Linux:
+                    if (!StringUtils.equals(
+                            RuntimeStackUtils.getJavaVersionFromRuntimeStack(result.getRuntimeStack()),
+                            this.getRuntime().getJavaVersionRaw()) ||
+                        !StringUtils.equals(
+                                RuntimeStackUtils.getWebContainerFromRuntimeStack(result.getRuntimeStack()),
+                                this.getRuntime().getWebContainerRaw())) {
+                        FieldUtils.writeField(configuration, "runtimeStack", null, true);
```
Flanker32, line 165: add set method in `WebAppConfiguration`?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
```diff
@@ -322,82 +369,78 @@ private WebAppConfiguration getRuntimeConfiguration(WebAppConfiguration configur
...
         // For project which package is jar, use java se runtime
         if (isJarProject()) {
-            return builder.runtimeStack(RuntimeStackUtils.getRuntimeStack(javaVersion));
+            return builder.runtimeStack(RuntimeStackUtils.getJavaSERuntimeStack(javaVersion));
+        }
+        final List<String> validRuntimeStacks = RuntimeStackUtils.getValidLinuxRuntimeStacksForJavaVersion(javaVersion);
+        String defaultLinuxRuntimeStack = configuration.getLinuxRuntimeStackOrDefault();
+        // sometimes the combination of <javaVersion> and <webContainer> may not be valid, but
+        // <webContainer> might be right
+        if (Objects.isNull(configuration.getRuntimeStack()) && Objects.nonNull(this.getRuntime()) &&
```
Flanker32, line 390: It may be a bit confused to show validation result here, we may show validation before `config` and only use correct default value here

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
```diff
@@ -322,82 +369,78 @@ private WebAppConfiguration getRuntimeConfiguration(WebAppConfiguration configur
...
+        }
+
+        final String javaVersionInput = queryer.assureInputFromUser(JAVA_VERSION, defaultJavaVersion,
+                validJavaVersions, String.format(COMMON_PROMPT, JAVA_VERSION, defaultJavaVersion));
+        final JavaVersion javaVersion = JavaVersionUtils.toAzureSdkJavaVersion(javaVersionInput);
         // For project which package is jar, use java se runtime
         if (isJarProject()) {
-            return builder.javaVersion(javaVersion).webContainer(getJavaSEWebContainer(javaVersion));
+            final WebContainer webContainer = WebContainerUtils.getJavaSEWebContainer();
+            if (Objects.isNull(webContainer)) {
```
Flanker32, line 425: `WebContainerUtils.getJavaSEWebContainer()` will return `WebContainer.fromString(JAVA_SE)` directly, will it be null sometime?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
```diff
@@ -543,7 +586,10 @@ private static WebAppConfiguration getConfigurationFromExisting(WebApp webapp, A
             builder.os(OperatingSystemEnum.Docker);
         } else if (webapp.operatingSystem() == OperatingSystem.LINUX) {
             builder.os(OperatingSystemEnum.Linux);
-            builder.runtimeStack(AppServiceUtils.parseRuntimeStack(webapp.linuxFxVersion()));
+            final RuntimeStack runtimeStack = AppServiceUtils.parseRuntimeStack(webapp.linuxFxVersion());
+            builder.runtimeStack(runtimeStack);
```
Flanker32, line 590: `RuntimeStack` should contains the java info, do we need to set `javaVersion` in builder?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
andxu: no need since it breaks the builder integration, this piece of code will be deleted in future refactor, since we need a real model representing user's configuration other than WebConfiguration which is mainly used in deploy goal.

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
andxu: If we don't show it here, user will be strange why his values are not replaced(if his value is illegal).

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/ConfigMojo.java
andxu: this code is copied from:

 private static WebContainer getJavaSEWebContainer(JavaVersion javaVersion) throws MojoFailureException {	
        final String version = javaVersion.toString();	
        if (StringUtils.startsWith(version, JAVA_1_8)) {	
            return WebContainer.JAVA_8;	
        } else if (StringUtils.startsWith(version, JAVA_11)) {	
            return WebContainer.fromString(JAVA_11_STRING);	
        } else {	
            throw new MojoFailureException(String.format("Java SE environment is not supported in %s", javaVersion.toString()));	
        }	
    }	    }


since the java version is removed from webContainer , it is OK to remove this error.

Flanker32 (approved):

PR #1206, opened 2021-01-05
add exception classes and AzureCredentialWrapper

andxu:


azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/tools/auth/model/AuthType.java
```diff
@@ -28,24 +30,30 @@ public static AuthType parseAuthType(String type) {
...
-            case "managed_identity":
-                return MANAGED_IDENTITY;
             case "oauth2":
                 return OAUTH2;
             case "visual_studio":
                 return VISUAL_STUDIO;
-            case "auto":
-                return AUTO;
+            default:
+                throw new UnsupportedOperationException(String.format("Invalid auth type '%s', supported values are: %s.", type,
```
Flanker32, line 54: use `IllegalArgumentException` here?

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/tools/auth/exception/AzureLoginException.java
```diff
@@ -0,0 +1,13 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
+package com.microsoft.azure.tools.auth.exception;
+
+public class AzureLoginException extends Exception {
```
Flanker32, line 9: Do we need to have a parent class like `AzureException`

andxu: this is auth lib, all auth exceptions should inherit this exception

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/tools/auth/model/AuthType.java
andxu: InvalidConfigurationException  will be better

Flanker32 (approved):
LGTM~

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/tools/auth/exception/LoginFailureException.java
```diff
@@ -0,0 +1,16 @@
...
+package com.microsoft.azure.tools.auth.exception;
+
+public class LoginFailureException extends AzureLoginException {
+    private static final long serialVersionUID = -94207672112213624L;
+
+    public LoginFailureException(String message) {
+        super(message);
+    }
+
+}
```
shenqianjin, line 16: Is it really necessary for us to create these new exceptions whose only difference is the class name?
how about only create one/two for us, let us use different values of its properties to describe different cases?

andxu: Rule 2:
image

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/tools/auth/model/AuthType.java
```diff
@@ -8,44 +8,53 @@
...
-            case "managed_identity":
-                return MANAGED_IDENTITY;
             case "oauth2":
                 return OAUTH2;
             case "visual_studio":
                 return VISUAL_STUDIO;
-            case "auto":
-                return AUTO;
+            default:
+                throw new InvalidConfigurationException(String.format("Invalid auth type '%s', supported values are: %s.", type,
```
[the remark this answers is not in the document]
andxu: This piece of code is running without costing too much memory and it merely runs, using a map(hash map) will cost permanent memory in heap.

Issue #844, opened 2019-09-30
[Feature] Support proxy when accessing Azure Resources


andxu:

andxu:
fixed in https://github.com/microsoft/azure-maven-plugins/pull/1240

PR #1225, opened 2021-01-18
[WIP]support proxy in maven plugins(exclude asc)

andxu:
Fix issue: https://github.com/microsoft/azure-maven-plugins/issues/844

### DIFF
--- a/azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/tools/auth/AzureAuthManager.java
+++ b/azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/tools/auth/AzureAuthManager.java
@@ -21,6 +21,7 @@ import com.microsoft.azure.tools.auth.exception.LoginFailureException;
 import com.microsoft.azure.tools.auth.model.AuthConfiguration;
 import com.microsoft.azure.tools.auth.model.AuthType;
 import com.microsoft.azure.tools.auth.model.AzureCredentialWrapper;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import rx.Single;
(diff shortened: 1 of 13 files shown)


azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/tools/auth/util/ValidationUtil.java
```diff
@@ -41,17 +41,18 @@ public static AuthConfiguration validateMavenAuthConfiguration(AuthConfiguration
     }
 
     public static void validateHttpProxy(String httpProxyHost, String httpProxyPort) throws InvalidConfigurationException {
-        if ((StringUtils.isNotBlank(httpProxyHost) && StringUtils.isBlank(httpProxyPort)) ||
-                (StringUtils.isBlank(httpProxyHost) && StringUtils.isNotBlank(httpProxyPort))) {
+        if (StringUtils.isAllBlank(httpProxyHost, httpProxyPort)) {
+            return;
+        }
+        if (!StringUtils.isNoneBlank(httpProxyHost, httpProxyPort)) {
```
Flanker32, line 47: May use `isAnyBlank(String...)` here

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/tools/auth/AzureAuthManager.java
```diff
@@ -66,7 +71,11 @@
         }
         return retrievers.retrieve().onErrorResumeNext(e ->
                 Single.error(new LoginFailureException(String.format("Cannot get credentials from authType '%s' due to error: %s", authType, e.getMessage())))
-        );
+        ).doOnSuccess(credentialWrapper -> {
+            if (StringUtils.isNotBlank(configuration.getHttpProxyHost())) {
```
Flanker32, line 75: Do we need to set proxy before retrieve as we need to validate the token which may need network access

andxu: seems we needn't to save proxy back to CredentialWrapper

andxu:
replaced with https://github.com/microsoft/azure-maven-plugins/pull/1240

PR #1249, opened 2021-01-22
support -D parameter for auth&runtime

andxu:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
```diff
@@ -339,6 +340,7 @@ protected Azure getOrCreateAzureClient() throws AzureAuthFailureException, Azure
         try {
             final MavenAuthConfiguration mavenAuthConfiguration = auth == null ? new MavenAuthConfiguration() : auth;
             mavenAuthConfiguration.setType(getAuthType());
```
wangmingliang-ms, line 342: this should be done immediately after the mojo properties are ready.

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/utils/SystemPropertyUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import java.lang.reflect.Modifier;
+import java.util.List;
+
+public class SystemPropertyUtils {
+    public static void injectCommandLineParameter(String prefix, Object obj, @Nonnull Class cls) {
+        Object result = obj;
+        try {
+            if (result == null) {
+                try {
+                    result = cls.getDeclaredConstructor().newInstance();
```
wangmingliang-ms, line 25: result is never used.

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
```diff
@@ -339,6 +340,7 @@ protected Azure getOrCreateAzureClient() throws AzureAuthFailureException, Azure
...
+            SystemPropertyUtils.injectCommandLineParameter("auth", mavenAuthConfiguration, MavenAuthConfiguration.class);
```
wangmingliang-ms, line 343: will `mavenAuthConfiguration` be null?
what happens if `mavenAuthConfiguration` is null, it seems your implementation does not support null.
and if yes, the third parameter is not needed.

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/utils/SystemPropertyUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.commons.lang3.reflect.FieldUtils;
+
+import javax.annotation.Nonnull;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Modifier;
+import java.util.List;
+
+public class SystemPropertyUtils {
+    public static void injectCommandLineParameter(String prefix, Object obj, @Nonnull Class cls) {
```
wangmingliang-ms, line 20: if `obj` is null, all the tasks done is this method is useless and affect nothing.

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
andxu: mavenAuthConfiguration should be always non-null, but the other parameter runtime may be null

wangmingliang-ms (approved):

Issue #1209, opened 2021-01-08
Use azure-maven-plugin to deploy python custom docker image


iyunbo:
While deploying custom docker image with python runtime,
the worker runtime is overwritten by java, due to 

https://github.com/microsoft/azure-maven-plugins/blob/5d8e6b9d5bc171b9b4b7570072b9bdb72fc0c434/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java#L99-L100

can you enable the deployment for python in case of  custom image? 

In fact, this light modification could help a lot on development of python custom image.

iyunbo:
any update please ?

Flanker32:
@iyunbo Thanks a lot for your advice, here is the pr for this issue https://github.com/microsoft/azure-maven-plugins/pull/1250

iyunbo:
Excellent news

Flanker32:
@iyunbo Thanks for your help and really sorry for the delay, functions maven plugin 1.10.0 has been released with this feature, you may get the change log [here](https://github.com/microsoft/azure-maven-plugins/releases/tag/azure-functions-maven-plugin-1.10.0)

PR #1250, opened 2021-01-24
Enable users set different functions worker with maven plugin

Flanker32:
Enable users set different functions worker with maven plugin, for #1209


azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
```diff
@@ -96,8 +96,8 @@
     //region get App Settings
     public Map getAppSettingsWithDefaultValue() {
         final Map settings = getAppSettings();
-        overrideDefaultAppSetting(settings, FUNCTIONS_WORKER_RUNTIME_NAME, SET_FUNCTIONS_WORKER_RUNTIME,
```
andxu, line 99: shall we remove the func: overrideDefaultAppSetting

andxu (changes requested):
reviewed with some comments

andxu (approved):

PR #1259, opened 2021-01-26
Print auth details for asc

andxu:
image


azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/springcloud/AbstractMojoBase.java
```diff
@@ -267,6 +272,12 @@ public SpringCloudAppConfig getConfiguration() {
     public AppPlatformManager getAppPlatformManager() {
         if (this.manager == null) {
             final LogLevel logLevel = getLog().isDebugEnabled() ? LogLevel.BODY_AND_HEADERS : LogLevel.NONE;
+            final AzureEnvironment env = azureCredentialWrapper.getEnv();
+            final String environmentName = AuthHelper.azureEnvironmentToString(env);
+            if (env != AzureEnvironment.AZURE) {
```
shenqianjin, line 277: AzureEnvironment is enum ?
I prefer to user !AzureEnvironment.AZURE.equals(env), no matter what type it it.

andxu: it is a standard constant field, like enum, there are no reasons to create another 'AZURE' enviromnent.

wangmingliang-ms (approved):

PR #1267, opened 2021-01-27
Get artifact by finalName in case use run azure-functions:package directly

Flanker32:
Get artifact by final name in case use run azure-functions:package directly


andxu (dismissed):

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
```diff
@@ -114,14 +114,15 @@ private void setDefaultAppSetting(Map result, String settingName, String setting
     }
 
     private void setDefaultAppSetting(Map result, String settingName, String settingIsEmptyMessage,
-                                        String settingValue, String warningMessage) {
+                                        String defaultValue, String warningMessage) {
```
andxu, line 122: var: setting should be replaced with value;

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -358,6 +369,42 @@ protected void checkArtifactCompileVersion() throws AzureExecutionException {
 
     protected void parseConfiguration() {
         parsedJavaVersion = FunctionUtils.parseJavaVersion(getRuntime().getJavaVersion());
+        appSettings = processAppSettingsWithDefaultValue();
+    }
+
+    public Properties processAppSettingsWithDefaultValue() {
```
andxu, line 375: no need to return, update appSettings

andxu (approved):

PR #1275, opened 2021-01-27
[WIP] App service common library with track 2

Flanker32:
- Implement app service common library with track 2
- Fix issues in webapp maven plugin


azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/service/impl/AppServiceUtils.java
```diff
@@ -173,4 +179,31 @@ public static AppServicePlanEntity getAppServicePlanEntity(com.azure.resourceman
                 .operatingSystem(getOperatingSystem(appServicePlan.operatingSystem()))
                 .build();
     }
+
+    static AppServicePlan getAppServicePlan(AppServicePlanEntity entity, AzureResourceManager azureClient) {
+        try {
+            return StringUtils.isNotEmpty(entity.getId()) ?
+                    azureClient.appServicePlans().getById(entity.getId()) :
+                    azureClient.appServicePlans().getByResourceGroup(entity.getResourceGroup(), entity.getName());
+        } catch (Exception e) {
```
andxu, line 188: change to an azure exception

andxu: for example: when the network is down, the exception will be thrown, then do you think this plan doesn't exist?

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/service/impl/AppServiceUtils.java
```diff
@@ -173,4 +179,31 @@ public static AppServicePlanEntity getAppServicePlanEntity(com.azure.resourceman
...
+            return null;
+        }
+    }
+
+    static DeployType getDeployTypeByFileExtension(File file) {
+        final String fileExtensionName = FilenameUtils.getExtension(file.getName());
+        switch (StringUtils.lowerCase(fileExtensionName)) {
+            case "jar":
+                return DeployType.JAR;
+            case "war":
```
andxu, line 199: DeployType.values().stream...

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/parser/V2ConfigParser.java
```diff
@@ -67,7 +67,7 @@ public DockerConfiguration getDockerConfiguration() throws AzureExecutionExcepti
     @Override
     public Runtime getRuntime() throws AzureExecutionException {
         final MavenRuntimeConfig runtime = mojo.getRuntime();
-        if (runtime == null) {
+        if (runtime == null || runtime.isEmpty()) {
```
shenqianjin, line 70: runtime.isEmpty() ? what's that ?

Flanker32: We will skip update runtime once user didn't set it. However, maven may auto create the `MavenRuntimeConfig` so that we need to use `isEmpty` to check whether user set the value

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/AzureAppService.java
```diff
@@ -79,7 +82,7 @@ public IWebAppDeploymentSlot deploymentSlot(String resourceGroup, String appName
     }
 
     public IWebAppDeploymentSlot deploymentSlot(WebAppDeploymentSlotEntity deploymentSlot) {
-        return null;
+        return new WebAppDeploymentSlot(deploymentSlot, this);
     }
 
     public AzureResourceManager getAzureResourceManager() {
```
shenqianjin, line 88: Some comments about this class:

NO. 1: renaming : AzureAppService --> AzureAppServiceHelper/AzureAppServiceUtil? actually, i think this class is only a static util class.

image

No. 2: As azureResourceManager property is only required by 2 method. how about to let is as a method parameter?
image

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/model/Runtime.java
```diff
@@ -22,8 +22,6 @@
 @AllArgsConstructor
 public class Runtime {
```
shenqianjin, line 23: rename to AzureRuntime ?  a little confused to java.lang.Runtime.

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/service/AbstractAppServiceCreator.java
```diff
@@ -0,0 +1,71 @@
...
+import com.microsoft.azure.toolkits.appservice.model.PricingTier;
+import com.microsoft.azure.toolkits.appservice.model.Runtime;
+import com.microsoft.azure.tools.common.model.Region;
+import lombok.Getter;
+
+import java.util.Map;
+import java.util.Optional;
+
+@Getter
+public abstract class AbstractAppServiceCreator<T> implements IAppServiceCreator<T> {
```
shenqianjin, line 19: Double confirm to do like this for "@Getter". It maybe not skip these public/static properties. I prefer to code it above these private properties we need one by one.

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/service/impl/AppServiceUtils.java
```diff
@@ -173,4 +179,31 @@ public static AppServicePlanEntity getAppServicePlanEntity(com.azure.resourceman
...
+                return DeployType.WAR;
+            case "ear":
+                return DeployType.EAR;
+            case "zip":
+                return DeployType.ZIP;
```
shenqianjin, line 204: how about to add a parse method in DeployType ?

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/service/impl/WebApp.java
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    @Override
+    public Runtime getRuntime() {
+        return AppServiceUtils.getRuntimeFromWebApp(getWebAppClient());
+    }
+
+    @Override
+    public WebAppUpdater update() {
+        return new WebAppUpdater();
```
shenqianjin, line 121: it is un-completed implementation ?

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/service/impl/WebApp.java
```diff
@@ -0,0 +1,302 @@
...
+
+    @Override
+    public String id() {
+        return getWebAppClient().id();
+    }
+
+    @Override
+    public String name() {
+        return getWebAppClient().name();
+    }
```
shenqianjin, line 166: is it necessary for us ? how about let entity() responsible for these basic properties ?

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/service/impl/WebApp.java
```diff
@@ -0,0 +1,302 @@
...
+                case WINDOWS:
+                    withCreate = createWindowsWebApp(blank, resourceGroup, appServicePlan, runtime);
+                    break;
+                case DOCKER:
+                    final DockerConfiguration dockerConfiguration = getDockerConfiguration().get();
+                    withCreate = createDockerWebApp(blank, resourceGroup, appServicePlan, dockerConfiguration);
+                    break;
+                default:
+                    throw new AzureToolkitRuntimeException(String.format(UNSUPPORTED_OPERATING_SYSTEM, runtime.getOperatingSystem()));
+            }
```
shenqianjin, line 190: I prefer to user if else if to replace switch code.

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/service/impl/WebAppDeploymentSlot.java
```diff
@@ -0,0 +1,201 @@
...
+                final WebApp webAppService = StringUtils.isNotEmpty(slotEntity.getId()) ?
+                        azureClient.webApps().getById(slotEntity.getId().substring(0, slotEntity.getId().indexOf("/slots"))) :
+                        azureClient.webApps().getByResourceGroup(slotEntity.getResourceGroup(), slotEntity.getWebappName());
+                deploymentSlotClient = StringUtils.isNotEmpty(slotEntity.getId()) ? webAppService.deploymentSlots().getById(slotEntity.getId()) :
+                        webAppService.deploymentSlots().getByName(slotEntity.getName());
+                slotEntity = AppServiceUtils.getWebAppDeploymentSlotEntity(deploymentSlotClient);
+            } catch (Exception e) {
+                // SDK will throw exception when resource not founded
+                return null;
+            }
```
shenqianjin, line 127: we do nothing for this exception ?

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/service/impl/WebApp.java
Flanker32: As most usage of entity is to get the name and id, so offer id and name in manager interface directly

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/service/impl/WebApp.java
Flanker32: No, return Updater in `update()` method, and commit by the commit method of Updater

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/service/impl/WebAppDeploymentSlot.java
Flanker32: Yes, we expect null for get a non-exists resource instead of exception

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/service/impl/WebApp.java
Flanker32: IMO switch may be more clear to handle different os in different ways

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkits/appservice/service/impl/AppServicePlan.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        @Override
+        public IAppServicePlanCreator withOperatingSystem(OperatingSystem operatingSystem) {
+            this.operatingSystem = operatingSystem;
+            return this;
+        }
+
+        @Override
+        public IAppServicePlan commit() {
```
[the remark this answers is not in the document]
Flanker32: We discussed before to use commit for both creator and updater

PR #1317, opened 2021-02-22
[identity][auth][wip]supports tenant in auth-lib

andxu:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
```diff
@@ -339,30 +339,41 @@ protected Azure getOrCreateAzureClient() throws AzureAuthFailureException, Azure
...
-            return AzureClientFactory.getAzureClient(azureCredentialWrapper, getUserAgent());
+
+            this.subscriptionId = targetSubscriptionId;
+            return AzureClientFactory.getAzureClient(getUserAgent(), targetSubscriptionId);
         } catch (AzureLoginException | IOException e) {
             throw new AzureAuthFailureException(e.getMessage());
         }
     }
 
+    protected void printCredentialDescription(Account account) {
```
Flanker32, line 364: Move to Account library to reduceduplicate codes?

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/AbstractWebAppMojo.java
```diff
@@ -423,96 +416,42 @@ protected AzureAppService getOrCreateAzureAppServiceClient() throws AzureExecuti
...
-            final AzureCredentialWrapper azureCredentialWrapper = MavenAuthManager.getInstance().login(session, settingsDecrypter, mavenAuthConfiguration);
-            if (Objects.isNull(azureCredentialWrapper)) {
+            MavenAuthManager.getInstance().login(MavenAuthManager.getInstance().buildAuthConfiguration(session, settingsDecrypter, mavenAuthConfiguration));
+            Account account = com.microsoft.azure.toolkit.lib.Azure.az(AzureAccount.class).account();
+            if (!account.isAuthenticated()) {
                 return null;
             }
-            final com.microsoft.azure.AzureEnvironment env = azureCredentialWrapper.getEnv();
+
+            final List<SubscriptionEntity> subscriptions = account.getSubscriptions();
```
Flanker32, line 425: extract codes below to `getSubscriptionId` and `promptAzureEnvironment`?

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/model/SubscriptionOption.java
```diff
@@ -35,12 +31,12 @@ public String toString() {
 
     @Override
     public int compareTo(SubscriptionOption other) {
-        final String name1 = inner != null ? inner.displayName() : null;
-        final String name2 = other.inner != null ? other.inner.displayName() : null;
+        final String name1 = inner != null ? inner.getName() : null;
```
Flanker32, line 34: Do we need to compare the subscription id here?

azure-spring-cloud-maven-plugin/src/main/java/com/microsoft/azure/maven/springcloud/AbstractMojoBase.java
```diff
@@ -146,26 +148,28 @@ public void execute() throws MojoFailureException {
...
     }
 
-    protected void initExecution() throws MojoFailureException, MavenDecryptException, AzureExecutionException,
-        com.microsoft.azure.toolkit.lib.auth.exception.InvalidConfigurationException {
+    protected void initExecution() throws MojoFailureException, MavenDecryptException, AzureExecutionException {
         // init proxy manager
         ProxyUtils.initProxy(Optional.ofNullable(this.session).map(s -> s.getRequest()).orElse(null));
         // Init telemetries
         initTelemetry();
+        Azure.az().config().setLogLevel(LogLevel.NONE);
```
Flanker32, line 156: Do we also need to add same codes to azure-maven-plugin-lib?

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -5,34 +5,176 @@
...
+                }
+
+            }
+        });
+
+        // since some tenants doesn't have subscriptions, reduce the tenants
+        this.entity.setTenantIds(new ArrayList<>(validTenantIds));
+        this.entity.setSubscriptions(new ArrayList<>(subscriptionMap.values()));
+        if (subscriptionMap.isEmpty()) {
+            this.entity.setAuthenticated(false);
```
Flanker32, line 99: For users with no subscription, should we set them as unauthenticated?

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -5,34 +5,176 @@
...
+
+                for (SubscriptionEntity s : subscriptionsOnTenant) {
+                    String key = StringUtils.lowerCase(s.getId());
+                    if (subscriptionMap.putIfAbsent(key, s) != null) {
+                        validTenantIds.add(tenantId);
+                    }
+                }
+            } catch (Exception ex) {
+                // ignore AuthenticationException since on some tenants, it doesn't allow list subscriptions
+                if (!(ExceptionUtils.getRootCause(ex) instanceof AuthenticationException)) {
```
Flanker32, line 87: for other exception, do we need to throw them out?

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -5,34 +5,176 @@
...
+        this.entity.setTenantIds(new ArrayList<>(validTenantIds));
+        this.entity.setSubscriptions(new ArrayList<>(subscriptionMap.values()));
+        if (subscriptionMap.isEmpty()) {
+            this.entity.setAuthenticated(false);
+        }
+    }
+
+    public void selectSubscriptions(List<String> subscriptionIds) {
+        // select subscriptions
+        if (CollectionUtils.isNotEmpty(subscriptionIds) && CollectionUtils.isNotEmpty(this.entity.getSubscriptions())) {
```
Flanker32, line 105: Do we need to set all subscriptions unselected to clear previous selection status?

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -5,34 +5,176 @@
...
+
+    public boolean isAuthenticated() {
+        return this.entity != null && this.entity.isAuthenticated();
+    }
+
+    public List<SubscriptionEntity> getSubscriptions() {
+        if (this.entity != null) {
+            return this.entity.getSubscriptions();
+        }
+        return null;
```
Flanker32, line 132: return empty list instead of null?

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -5,34 +5,176 @@
...
+            return this.entity.getSelectedSubscriptions();
+        }
+        return null;
+    }
+
+    public TokenCredential getCredential(String subscriptionId) throws LoginFailureException {
+        if (!this.isAuthenticated()) {
+            throw new LoginFailureException("Please login first.");
+        }
+        Objects.requireNonNull(this.credentialBuilder, "Azure Account should be initialized first.");
```
Flanker32, line 146: Change messages to login first? As seems we will not do nothing for credentialBuilder in `initialize()`

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/AzureAccount.java
```diff
@@ -5,38 +5,135 @@
...
+        this.account = new Account();
+        final AccountEntity entity = new AccountEntity();
+        entity.setMethod(accountEntity.getMethod());
+        entity.setAuthenticated(accountEntity.isAuthenticated());
+        entity.setEnvironment(accountEntity.getEnvironment());
+        entity.setEmail(accountEntity.getEmail());
+        entity.setTenantIds(accountEntity.getTenantIds());
+        this.account.setEntity(entity);
+        this.account.setCredentialBuilder(accountEntity.getCredentialBuilder());
+        if (!accountEntity.isAuthenticated()) {
```
Flanker32, line 102: Check whether account entity is authenticated first?

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/AzureAccount.java
```diff
@@ -5,38 +5,135 @@
...
+            if (loginByBuilder(builder)) {
+                return;
+            }
+        }
+    }
+
+    public void login(AccountEntity accountEntity) throws AzureToolkitAuthenticationException {
+        Objects.requireNonNull(accountEntity, "accountEntity is required.");
+        this.account = new Account();
+        final AccountEntity entity = new AccountEntity();
```
Flanker32, line 94: create a copy constructor for `AccountEntity`?

azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
andxu: Good idea

andxu: The only concern is here I need a common ILog or Messager for prompting something, since log levels in maven and gradle are different

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/AzureAccount.java
andxu: We cannot reuse the account entity because it is the inner state of Current AzureAccount => Account, otherwise the update on accountEnity passed to login will impact the the inner state AzureAccount

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/AzureAccount.java
andxu: Login is likely to change the state(Auth method) of AzureAccount, if login fails, the account is left with the failure login with
authenticated = false

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
andxu: It is fine to have null credentialBuilder when login fails

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
andxu: if the entity is null, we should return null, this is a shortcut for accountEntity on AzureAccount

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
andxu: selectSubscriptions should be private can be only called once and cannot be called again.

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
andxu: Let it throw exception and when user reports a valid exception, we should allow it here.

PR #1320, opened 2021-02-23
[asc, sharecode]fix to intellij's requirements

wangmingliang-ms:


azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/model/IArtifact.java
```diff
@@ -0,0 +1,51 @@
...
+
+package com.microsoft.azure.toolkit.lib.common.model;
+
+import lombok.Getter;
+
+import javax.annotation.Nonnull;
+import java.io.File;
+
+public interface IArtifact {
+    String getId();
```
shenqianjin, line 15: why not a super POJO ? I feel that "id" is only a property of a object.

wangmingliang-ms: artifact is not a azure resource like `mysql` or `webapp`/`function`. 
and I dont see the necessity to abstract them at least for now and we may do this abstraction in the future when needed.

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/model/IArtifact.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Nonnull
+    static FileArtifact fromFile(@Nonnull final File file) {
```
Flanker32, line 18: IMO, library just need to define the Artifact interface and tooling will provide the imlpements as well as the from method

Flanker32: Besides, move this method to `FileArtifact`? I think we should not use the implementation of interface in interface itself.

azure-toolkit-libs/azure-toolkit-springcloud-lib/src/main/java/com/microsoft/azure/toolkit/lib/springcloud/config/SpringCloudAppConfig.java
```diff
@@ -6,22 +6,30 @@
...
+import lombok.NoArgsConstructor;
+import lombok.Setter;
 import org.apache.commons.lang3.BooleanUtils;
 
-@Getter
 @Builder
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
```
Flanker32, line 20: Why we need both `NoArgsConstructor` and `AllArgsConstructor` ?

PR #1335, opened 2021-03-02
[Azure-Identity][Refactor] Code refactor

andxu:
1. add sp credentials
2. rename MasterTokenCredential
3. delete some useless class
4. add cached token credential


azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -52,14 +62,24 @@ public TokenCredential getTokenCredential(String tenantId) {
...
     public AzureTokenCredentials getTokenCredentialV1(String tenantId) {
         return AzureTokenCredentialsAdapter.from(getEnvironment(), tenantId, getTokenCredential(tenantId));
     }
 
+    public TokenCredential getTokenCredentialForSubscription(String subscriptionId) {
+        SubscriptionEntity subscriptionEntity = getSubscriptionById(subscriptionId);
+        return getTokenCredential(subscriptionEntity.getTenantId());
+    }
+
+    public AzureTokenCredentials getTokenCredentialV1ForSubscription(String subscriptionId) {
```
wangmingliang-ms, line 78: the API should be clean(only depends on identity2) can just provide the adapater(in standalone package).
let the API caller to adapt `TokenCredential` to `AzureTokenCredential`

andxu: TokenCredential cannot be converted to AzureTokenCredential, it will need additional info such as 'env' and 'tenant id', if we move another package, the code will not be clean, just keep it here now, and in future, it will be deleted

wangmingliang-ms (approved):

PR #1350, opened 2021-03-12
Limit to selected subscriptions scope on account 

andxu:
1. Check isAuthenticated for methods which require authenticated
2. Get credential only for selected subscriptions
3. Throw error when no subscription is selected.

### DIFF
--- a/azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
+++ b/azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
@@ -60,6 +60,7 @@ public abstract class Account implements IAccount {
     }
 
     public TokenCredential getTokenCredential(String tenantId) {
+        requireAuthenticated();
         if (StringUtils.isBlank(tenantId)) {
             return this.entity.getCredential();
         } else {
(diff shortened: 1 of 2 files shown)


azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -68,16 +69,19 @@ public TokenCredential getTokenCredential(String tenantId) {
     }
 
     public AzureTokenCredentials getTokenCredentialV1(String tenantId) {
+        requireAuthenticated();
         return AzureTokenCredentialsAdapter.from(getEnvironment(), tenantId, getTokenCredential(tenantId));
     }
 
     public TokenCredential getTokenCredentialForSubscription(String subscriptionId) {
-        Subscription subscription = getSubscriptionById(subscriptionId);
+        requireAuthenticated();
```
wangmingliang-ms, line 77: I would suggest to rename 
`getTokenCredentialForSubscription` -> `getTokenCredential`
`getTokenCredential` -> `getTokenCredentialForTenant`
since the user would use `getTokenCredential` of subscription more frequently.

wangmingliang-ms (approved):

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
andxu: Will be rename later

PR #1352, opened 2021-03-16
bugfix: anonymous  telemetry are sent when error occurs in name-not-specified task.

wangmingliang-ms:


azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/operation/AzureOperationAspect.java
```diff
@@ -28,38 +28,31 @@ public void operation() {
...
+        final AzureOperationRef current = toOperationRef(point);
+        final AzureOperationRef operation = (AzureOperationRef) AzureTaskContext.current().popOperation();
+        AzureTelemeter.afterExit(operation);
+        assert Objects.equals(current, operation) : String.format("popped operation[%s] is not the exiting operation[%s]", current, operation);
     }
 
     @AfterThrowing(pointcut = "operation()", throwing = "e")
     public void afterThrowing(JoinPoint point, Throwable e) throws Throwable {
-        final AzureOperationRef operation = exitOperation(point);
+        final AzureOperationRef current = toOperationRef(point);
```
Flanker32, line 46: the validation seems duplicate with `afterReturning` (except the assert message), can we share the logic?

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/operation/AzureOperationAspect.java
```diff
@@ -28,38 +28,31 @@ public void operation() {
...
+        final AzureOperationRef operation = (AzureOperationRef) AzureTaskContext.current().popOperation();
+        AzureTelemeter.onError(operation, e);
```
Flanker32, line 48: Do we need to remove other `AzureTelemeter.onError` call, like codes in `AzureTaskContext` in case duplicated?

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/operation/AzureOperationAspect.java
wangmingliang-ms: no, actually there are only two-lines same code. sharing code will make the code ugly.

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/operation/AzureOperationAspect.java
wangmingliang-ms: no, they are not duplicated, `onError` in `AzureTaskContext` is only for `AzureTask`. this is for `AzureOperation` only.

Flanker32 (approved):

PR #1353, opened 2021-03-16
Fix authentication issue for web app maven plugin with new account library

Flanker32:
- Extract `getAzureAccount` from `getAzureClient` and share with `getOrCreateAzureAppServiceClient`
- Add missing subscription id in configuration parser
- Fix NPE in auth/app service library

### DIFF
--- a/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/AbstractWebAppMojo.java
+++ b/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/AbstractWebAppMojo.java
@@ -7,6 +7,7 @@ package com.microsoft.azure.maven.webapp;
 
 import com.microsoft.azure.common.appservice.DockerImageType;
 import com.microsoft.azure.common.exceptions.AzureExecutionException;
+import com.microsoft.azure.common.logging.Log;
 import com.microsoft.azure.common.utils.AppServiceUtils;
 import com.microsoft.azure.management.appservice.DeploymentSlot;
 import com.microsoft.azure.management.appservice.WebApp;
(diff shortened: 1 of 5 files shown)


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/AbstractWebAppMojo.java
```diff
@@ -399,8 +407,36 @@ protected WebAppConfig getWebAppConfig() throws AzureExecutionException {
...
     }
 
-    protected AzureAppService getOrCreateAzureAppServiceClient() {
-        return Azure.az(AzureAppService.class);
+    protected AzureAppService getOrCreateAzureAppServiceClient() throws AzureExecutionException {
+        if (appServiceClient == null) {
+            try {
+                final Account account = getAzureAccount();
+                final List<Subscription> subscriptions = account.getSubscriptions();
+                final String targetSubscriptionId = getTargetSubscriptionId(getSubscriptionId(), subscriptions, account.getSelectedSubscriptions());
```
andxu, line 415: I think we should move the selection of one target subscription in getAzureAccount by calling select subscription, then we will not need to have the logic like:

appServiceClient == null ? this.subscriptionId : appServiceClient.getDefaultSubscription().getId();

public String getSubscriptionId() {
        return mojo.getSubscriptionId();
    }

Flanker32: Yes, for command line tooling, as there will always be one subscription, we may select it in Account and always use the selected one, this should be done in later PR for all command line tooling.

andxu (approved):

PR #1357, opened 2021-03-19
cache warmup abstraction with @Preload

wangmingliang-ms:
```
    @Preload
    @Cacheable(value = "sdk-services", condition = "$force&&$force[0]")
    public static List<AzureSdkServiceEntity> loadAzureSdkServices(Boolean... force) throws IOException {
        return ...
    }
```

### DIFF
--- a/azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/cache/Preloader.java
+++ b/azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/cache/Preloader.java
@@ -22,8 +22,11 @@ import java.util.logging.Level;
 
 @Log
 public class Preloader {
+
+    private static final String INVALID_PRELOAD_METHOD = "@Preload annotated method(%s.%s) should have (no args or only varargs) " +
+            "and must be (static or in a singleton class)";
+
     public static Collection<Method> load() {
-        // Using progress manager as azure task manager is not initialized
         final Set<Method> methods = getPreloadingMethods();
         log.log(Level.INFO, String.format("Found %d @Preload annotated methods.", methods.size()));
         methods.parallelStream().forEach((m) -> {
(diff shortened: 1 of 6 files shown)


azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/cache/Preloader.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    @Nullable
+    private static Object getSingleton(final Method m) {
+        final Class<?> clazz = m.getDeclaringClass();
+        try {
+            final Method getInstance = clazz.getDeclaredMethod("getInstance");
+            if (Modifier.isStatic(getInstance.getModifiers()) && getInstance.getParameterCount() == 0) {
+                getInstance.setAccessible(true);
```
andxu, line 61: getInstance shall not be non public?

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/cache/Preloader.java
```diff
@@ -0,0 +1,77 @@
...
+        final Set<Method> methods = getPreloadingMethods();
+        log.log(Level.INFO, String.format("Found %d @Preload annotated methods.", methods.size()));
+        methods.parallelStream().forEach((m) -> {
+            Object instance = null;
+            // TODO: maybe support predefined variables, e.g. selected subscriptions
+            if ((m.getParameterCount() == 0 || m.isVarArgs())
+                    && (Modifier.isStatic(m.getModifiers()) || Objects.nonNull(instance = getSingleton(m)))) {
+                invoke(m, instance);
+            } else {
+                log.warning("@Preload annotated method should have (no args or only varargs) and must be (static or in a singleton class)");
```
andxu, line 36: add the class and method name in the error message?

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/cache/Preloader.java
wangmingliang-ms: use `setAccessible(true)` to guarantee this call will never fail.

PR #1358, opened 2021-03-19
Fix a few bugs

andxu:
NPE of to string when method is null
@Getter doesn't override getMethod

### DIFF
--- a/azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
+++ b/azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
@@ -130,10 +130,13 @@ public abstract class Account implements IAccount {
     @Override
     public String toString() {
         final List<String> details = new ArrayList<>();
-        if (getEntity() == null) {
+
+        if (this.isAuthenticated()) {
             return "<account not logged in>";
         }
-        details.add(String.format("Auth method: %s", TextUtils.cyan(getEntity().getMethod().toString())));
+        if (getMethod() != null) {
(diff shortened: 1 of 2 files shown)


azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -130,10 +130,13 @@ public void selectSubscription(List<String> selectedSubscriptionIds) {
     @Override
     public String toString() {
         final List<String> details = new ArrayList<>();
-        if (getEntity() == null) {
+
+        if (this.isAuthenticated()) {
```
Flanker32, line 134: should be `(!this.isAuthenticated())`?

Flanker32 (changes requested):

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -130,10 +130,13 @@ public void selectSubscription(List<String> selectedSubscriptionIds) {
...
             return "<account not logged in>";
         }
-        details.add(String.format("Auth method: %s", TextUtils.cyan(getEntity().getMethod().toString())));
+        if (getMethod() != null) {
```
Flanker32, line 137: Will we remove the check after all codes ready? When will we signed in without method?

andxu: No, account is abstract, we cannot ensure all account sub types will contain method

Flanker32 (approved):

wangmingliang-ms (approved):

PR #1372, opened 2021-03-25
evict cache with annotation @CacheEvict

wangmingliang-ms:


azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/cache/CacheManager.java
```diff
@@ -58,20 +65,59 @@ public Object around(ProceedingJoinPoint point) throws Throwable {
...
+        }
+        return point.proceed();
+    }
+
+    private void invalidateCache(@Nullable final String name, @Nullable final String key) throws ExecutionException {
+        if (StringUtils.isBlank(name)) { // invalidate all cache entries if cache name not specified
+            log.log(Level.INFO, "invalidate all cache");
+            caches.invalidateAll();
+        } else {
+            final Cache<Object, Object> cache = caches.get(name);
```
Flanker32, line 103: will cache be null a wrong cache name is provided?

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/cache/CacheManager.java
```diff
@@ -58,20 +65,59 @@ public Object around(ProceedingJoinPoint point) throws Throwable {
...
+            log.log(Level.INFO, String.format("evict cache[%s.%s] on method[%s]", name, key, signature.getName()));
+            invalidateCache(name, key);
+        }
+        return point.proceed();
+    }
+
+    private void invalidateCache(@Nullable final String name, @Nullable final String key) throws ExecutionException {
+        if (StringUtils.isBlank(name)) { // invalidate all cache entries if cache name not specified
+            log.log(Level.INFO, "invalidate all cache");
+            caches.invalidateAll();
```
Flanker32, line 101: Now the default value for `cacheName` is empty, which mean will invalidate all cache by default, which may be used by mistake. And I think it's a pretty dangerous operation to clear all cache, I prefer to user a keyword for all cache or provide a static method to do this job.  Same for clean all value in specific cache below.

PR #1387, opened 2021-03-31
Prepare release functions 1.10.1

Flanker32:
Prepare release functions 1.10.1


azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
```diff
@@ -131,6 +132,12 @@ public RuntimeConfiguration getRuntime() {
         return runtime;
     }
 
+    protected void validateAppName() {
+        if (StringUtils.isEmpty(appName)) {
```
andxu, line 136: is blank

PR #1391, opened 2021-04-07
Add async methods for auth-lib

andxu:


azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -5,66 +5,82 @@
...
-        return isAvailable();
-    }
+    protected abstract TokenCredential createTokenCredential();
 
     public AzureEnvironment getEnvironment() {
         return entity == null ? null : entity.getEnvironment();
     }
 
+    boolean checkAvailable() {
+        if (this.entity.isAvailable()) {
```
Flanker32, line 71: Why we refresh the available status only if it is not available? Will it always be true once available?

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -95,7 +111,65 @@ public boolean isAvailable() {
...
+            return this;
+        }).flatMap(ignore -> {
+            if (this.entity.getSubscriptions() == null) {
+                return initializeSubscriptions().then(Mono.just(this));
+            }
+            return Mono.just(this);
+        });
+    }
+
+    private String getEmailAndRefreshToken(AccessToken accessToken) {
```
Flanker32, line 142: Shall we split the method into `refershEmail` and `getRefreshToken`? As now it do return the refresh token, but modify the email address with side effect.

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -95,7 +111,65 @@ public boolean isAvailable() {
...
+    private Mono<Tuple2<List<String>, AccessToken>> verifyAndGetTenantId(TokenCredential credential) {
+        AzureEnvironment env = ObjectUtils.firstNonNull(this.entity.getEnvironment(), Azure.az(AzureCloud.class).getOrDefault());
+        TokenRequestContext tokenRequestContext = new TokenRequestContext()
+                .addScopes(ScopeUtil.resourceToScopes(env.getManagementEndpoint()));
+        Mono<AccessToken> mono = credential instanceof TenantCredential ?
+                ((TenantCredential) credential).getAccessToken(null, tokenRequestContext) :
+                credential.getToken(tokenRequestContext);
+
+        return mono.flatMap(accessToken ->
+                Mono.zip(AzureResourceManager.authenticate(request -> Mono.just(accessToken), new AzureProfile(env))
```
Flanker32, line 171: Please refer https://github.com/microsoft/azure-maven-plugins/blob/feature-auth/azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/AzureAppService.java#L135 to set the user agent with track 2 SDK, we may need to do this in the common library

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
andxu: for avail: one check and then remain unchanged

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
andxu: there has shared code 
```
        IAuthenticationResult result = ((MsalToken) accessToken).getAuthenticationResult();
        if (result == null) {
            return null;
        }

```

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/AzureAccount.java
```diff
@@ -46,129 +49,140 @@ public Account account() throws AzureToolkitAuthenticationException {
...
 
-    public AzureAccount login(AuthType type) throws LoginFailureException {
-        AuthConfiguration auth = new AuthConfiguration();
-        auth.setType(type);
-        loginWithAuthConfiguration(auth);
-        return this;
+    public Mono<Account> authenticateAccounts(List<Account> accounts) {
+        Mono<Account> current = accounts.get(0).authenticateAsync();
+        for (int i = 1; i < accounts.size(); i++) {
+            current = current.onErrorResume(e -> accounts.get(0).authenticateAsync());
```
Flanker32, line 82: -> `e -> accounts.get(i).authenticateAsync()` ?

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
andxu: user agent should be in global config, let's add a todo comment here(just keep this PR not far away from auth)

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/AzureAccount.java
andxu: good catch

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
Flanker32: Yes, but it is a bit confused that one get method for both email and refresh token

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -95,7 +111,65 @@ public boolean isAvailable() {
...
+            refreshTokenFromResult = (String) FieldUtils.readField(result, "refreshToken", true);
+        } catch (IllegalAccessException e) {
+            throw new AzureToolkitAuthenticationException("Cannot read refreshToken from IAuthenticationResult.");
+        }
+
+        return refreshTokenFromResult;
+    }
+
+    private Mono<Tuple2<List<String>, AccessToken>> verifyAndGetTenantId(TokenCredential credential) {
+        AzureEnvironment env = ObjectUtils.firstNonNull(this.entity.getEnvironment(), Azure.az(AzureCloud.class).getOrDefault());
```
Flanker32, line 163: Do we need to handle the issue that entity environment differs from `AzureAccount` value?

andxu: in some cases, the env is from auth method, like azure cli, vscode

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
```diff
@@ -112,7 +186,7 @@ public boolean isAuthenticated() {
     public void selectSubscription(List<String> selectedSubscriptionIds) {
         requireAuthenticated();
         if (CollectionUtils.isEmpty(selectedSubscriptionIds)) {
-            throw new IllegalArgumentException("Cannot select subscription by an empty id list.");
+            throw new IllegalArgumentException("Cannot select subscription by empty subscription id list.");
```
Flanker32, line 189: `You must select at least one subscription`?

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/Account.java
Flanker32: In that case, should we change the account env value after authenticated?

PR #1416, opened 2021-04-16
[Bug] fix jdk 11 and netty dns issue

andxu:
1. replace http client with okhttp3 to fix netty java 11 issue
2. add global http proxy configuration and set proxy in http client

### DIFF
--- a/azure-toolkit-libs/pom.xml
+++ b/azure-toolkit-libs/pom.xml
@@ -51,12 +51,14 @@
 
         <azure.toolkit-lib.version>${project.version}</azure.toolkit-lib.version>
 
+        <azure.core.version>1.15.0</azure.core.version>
         <azure.aad.version>1.6.4</azure.aad.version>
         <azure.applicationinsights.version>2.6.2</azure.applicationinsights.version>
         <azure.applicationinsights-web.version>[2.0,)</azure.applicationinsights-web.version>
         <azure.azure-mgmt-appplatform.version>1.0.0-beta</azure.azure-mgmt-appplatform.version>
         <azure.client.version>1.7.10</azure.client.version>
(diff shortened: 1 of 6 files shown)


azure-toolkit-libs/pom.xml
```diff
@@ -257,8 +257,36 @@
...
+            </dependency>
+            <dependency>
+                <groupId>com.azure</groupId>
+                <artifactId>azure-core-http-okhttp</artifactId>
+                <version>1.6.1</version>
+            </dependency>
+            <dependency>
+                <groupId>com.squareup.okhttp3</groupId>
+                <artifactId>okhttp</artifactId>
+                <version>4.9.1</version>
```
wangmingliang-ms, line 288: make sure the other modules in this project use same version
make sure intellij plugin use same version.

azure-toolkit-libs/pom.xml
```diff
@@ -80,7 +80,7 @@
         <json.schema.validator.version>2.2.14</json.schema.validator.version>
         <groovy.version>2.5.11</groovy.version>
         <junit.version>4.13.1</junit.version>
-        <lombok.version>1.18.16</lombok.version>
+        <lombok.version>1.18.18</lombok.version>
```
wangmingliang-ms, line 83: why? need to prevent version conflicts with other modules and intellij plugin

andxu: intellij lombok plugin needs this change, otherwise the lombok java will not compile

azure-toolkit-libs/pom.xml
andxu: maven have add the new dependency, intellij will have update to this version in feature-auth branch

wangmingliang-ms (approved):

PR #1420, opened 2021-04-16
Support log streaming and related features in app service library

Flanker32:
- Add log stream interface in AppService
- Support get diagnostic config for app service
- Support deployment slot update, to enable log streaming
- Add entity for log level in diagnostic config


azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/model/LogLevel.java
```diff
@@ -0,0 +1,39 @@
...
+@Getter
+@RequiredArgsConstructor
+public class LogLevel {
+    public static final LogLevel OFF = new LogLevel("Off");
+    public static final LogLevel VERBOSE = new LogLevel("Verbose");
+    public static final LogLevel INFORMATION = new LogLevel("Information");
+    public static final LogLevel WARNING = new LogLevel("Warning");
+    public static final LogLevel ERROR = new LogLevel("Error");
+
+    private String value;
```
wangmingliang-ms, line 24: the constructor can be removed if this field is `final`

andxu (approved):

PR #1426, opened 2021-04-20
Copy resource in package mojo instead of maven resource/dependency plugin

Flanker32:
Copy resource in package mojo instead of maven resource/dependency plugin


andxu (commented):
Should share the same code with https://github.com/microsoft/azure-gradle-plugins/blob/master/azure-functions-gradle-plugin/src/main/java/com/microsoft/azure/plugin/functions/gradle/handler/PackageHandler.java

andxu (approved):

PR #1437, opened 2021-04-22
allow user to set telemetry properties directly from parameters.

wangmingliang-ms:
image


andxu (approved):

andxu:
Property is too common, TelemetryProperty?

wangmingliang-ms:
> Property is too common, TelemetryProperty?
user can actually use
`@AzureTelemetry.Property`
`@AzureTelemetry.Properties`

PR #1448, opened 2021-04-25
Fix the bugs: user config subscription is not selected

andxu:
1. Fix the bug : when user select A subs in azure cli but use B subs in configuration
2. Fix the bug : don't prompt current subscription used in spring cloud plugin
3. Fix the bug : the subscription id in telemetry of spring cloud plugin
4. Refactor:  remove duplicate code in AbstractAzureMojo and AbstractMojoBase


andxu:
replaced by #1459

PR #1468, opened 2021-05-06
[AzureMessager #2]azure messager framework

wangmingliang-ms:


azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/messager/IAzureMessager.java
```diff
@@ -0,0 +1,32 @@
...
+    default void warning(@Nonnull String message, String... title) {
+    }
+
+    default void error(@Nonnull String message, String... title) {
+    }
+
+    default void error(@Nonnull Throwable throwable, String... title) {
+    }
+
+    default String value(String val) {
```
andxu, line 29: what's this for?

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/messager/AzureMessager.java
```diff
@@ -0,0 +1,86 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Objects;
+import java.util.Optional;
+
+public abstract class AzureMessager implements IAzureMessager {
+    private static IAzureMessager defaultMessager;
+
+    public static synchronized void setDefaultMessager(@Nonnull IAzureMessager messager) {
+        if (AzureMessager.defaultMessager == null) {
```
andxu, line 21: how to unset the messager?

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/messager/AzureMessager.java
```diff
@@ -0,0 +1,86 @@
...
+            AzureMessager.defaultMessager = messager;
+        } else {
+            // not allow overwriting...
+            throw new AzureToolkitRuntimeException("default messager has already been registered");
+        }
+    }
+
+    @Nonnull
+    public static IAzureMessager getDefaultMessager() {
```
andxu, line 30: I think it is not supposed to be called from outside, make it private?

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/messager/IAzureMessager.java
```diff
@@ -0,0 +1,32 @@
+package com.microsoft.azure.toolkit.lib.common.messager;
+
+import javax.annotation.Nonnull;
+
+public interface IAzureMessager {
+
+    default boolean confirm(@Nonnull String message, String... title) {
```
andxu, line 7: beside boolean confirm, I think there might be some other input prompt like:
select subscription
select resource group
select runtime

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/messager/AzureMessager.java
wangmingliang-ms: this is the default messager, there is no need to unset.

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/messager/IAzureMessager.java
wangmingliang-ms: to wrap value so that value will be highlighted, e.g. in Maven, it would be `TextUtils.cyan(val)`

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/messager/AzureMessager.java
wangmingliang-ms: I am not sure whether we need to use the default messager directly.

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/messager/IAzureMessager.java
wangmingliang-ms: I don't think so. messager is only for **output** but `prompt` is for **input**.

andxu (approved):

PR #1470, opened 2021-05-06
[AzureMessager #3]set AzureMavenMessager as the default azure messager

wangmingliang-ms:


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/toolkit/maven/common/messager/AzureMavenMessager.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public void error(@Nonnull String message, @Nonnull String... title) {
+        log.error(message);
+    }
+
+    @Override
+    public void error(@Nonnull Throwable throwable, String... title) {
+        final String message = Objects.nonNull(title) ? title[0] : throwable.getMessage();
```
andxu, line 39: title[0] is not safe,  will throw IndexOutOfBoundsException

wangmingliang-ms: > title[0] is not safe, will throw IndexOutOfBoundsException

can we pass a `new String[0]` to `String...` varargs?

Issue #1473, opened 2021-05-06
Unknown value of the pricingTier for P1v3 P2v3 P3v3 (Linux)


Xeyor:
### Plugin name and version
com.microsoft.azure:azure-webapp-maven-plugin:1.14.0:deploy

### Plugin configuration in your `pom.xml`
```
<plugins>
	<plugin>
		<groupId>com.microsoft.azure</groupId>
		<artifactId>azure-webapp-maven-plugin</artifactId>
		<version>1.14.0</version>
		<configuration>
			<auth>
				<serverId>azure-auth</serverId>
			</auth>
			<schemaVersion>v2</schemaVersion>
			<subscriptionId>***************</subscriptionId>
			<resourceGroup>bkf-rg-dev</resourceGroup>
			<appName>bkf-int-api</appName>
			<pricingTier>P2v3</pricingTier>
			<region>westeurope</region>
			<runtime>
				<os>Linux</os>
				<javaVersion>Java 8</javaVersion>
				<webContainer>Java SE</webContainer>
			</runtime>
			<stopAppDuringDeployment>true</stopAppDuringDeployment>
			<deployment>
				<resources>
					<resource>
						<directory>${project.basedir}/target</directory>
						<includes>
							<include>*.jar</include>
						</includes>
					</resource>
				</resources>
			</deployment>
		</configuration>
	</plugin>
</plugins>
```

### Expected behavior
Deployment success to an existing PXv3 Linux AppService

### Actual behavior

```
[ERROR] Failed to execute goal com.microsoft.azure:azure-webapp-maven-plugin:1.14.0:deploy (default-cli) on project *****: Unknown value of the pricingTier. -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal com.microsoft.azure:azure-webapp-maven-plugin:1.14.0:deploy (default-cli) on project *****: Unknown value of the pricingTier.
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
[...]

PR #1474, opened 2021-05-07
Fix validation for pricingTier to support v3 pricing

Flanker32:
- Fix validation for pricingTier to support v3 pricing
- Add test cases for pricingTier parser

for issue https://github.com/microsoft/azure-maven-plugins/issues/1473

### DIFF
--- a/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/validator/AbstractConfigurationValidator.java
+++ b/azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/validator/AbstractConfigurationValidator.java
@@ -5,7 +5,7 @@
 
 package com.microsoft.azure.maven.webapp.validator;
 
-import com.microsoft.azure.toolkit.lib.legacy.appservice.AppServiceUtils;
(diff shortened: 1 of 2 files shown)


azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/validator/AbstractConfigurationValidator.java
```diff
@@ -48,7 +48,7 @@ public String validateResourceGroup() {
     }
 
     public String validatePricingTier() {
-        if (mojo.getPricingTier() != null && AppServiceUtils.getPricingTierFromString(mojo.getPricingTier()) == null) {
```
andxu, line 51: there are some more usages of getPricingTierFromString, shall we replace all of them?

andxu (approved):

azure-webapp-maven-plugin/src/main/java/com/microsoft/azure/maven/webapp/validator/AbstractConfigurationValidator.java
Flanker32: The method will be finally depredated after maven plugin track2 migration done

PR #1485, opened 2021-05-10
[Track2-Kudu-3] Support kudu operations in app service library

Flanker32:
- Support kudu operations in app service library
- Add `AppServiceFileLegacy` for compatibility (Will be removed after function track2 migration)


azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/service/IKuduClient.java
```diff
@@ -0,0 +1,36 @@
...
+import com.microsoft.azure.toolkit.lib.appservice.model.ProcessInfo;
+import com.microsoft.azure.toolkit.lib.appservice.model.TunnelStatus;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import java.nio.ByteBuffer;
+import java.util.List;
+
+public interface IKuduClient {
+    Flux<ByteBuffer> getFileContent(final String path);
```
andxu, line 19: why we change from  Mono<byte[]>
 to Flux<ByteBuffer>?

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/manager/AppServiceKuduManager.java
```diff
@@ -113,16 +130,27 @@ public void deleteFile(String path) {
...
+        final Mono<Response<TunnelStatus>> responseMono = this.kuduService.getAppServiceTunnelStatus(host);
+        return getValueFromResponseMono(responseMono);
+    }
+
+    private <T> T getValueFromResponseMono(Mono<Response<T>> response) {
+        return getValueFromResponseMono(response, null);
+    }
+
+    private <T> T getValueFromResponseMono(Mono<Response<T>> response, T defaultValue) {
+        return Optional.ofNullable(response.block()).map(Response::getValue).orElse(defaultValue);
```
andxu, line 153: Mono.block will never return null, so two getValueFromResponseMono methods are the same.

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/manager/AppServiceKuduManager.java
```diff
@@ -84,20 +92,29 @@ public static AppServiceKuduManager getClient(@Nonnull WebAppBase webAppBase) {
...
+        final Mono<Response<List<AppServiceFile>>> filesInDirectory = this.kuduService.getFilesInDirectory(host, dir);
+        return getValueFromResponseMono(filesInDirectory, Collections.emptyList()).stream()
                 .filter(file -> !"text/xml".equals(file.getMime()) || !file.getName().contains("LogFiles-kudu-trace_pending.xml"))
+                .map(file -> file.withApp(appService).withPath(Paths.get(dir, file.getName()).toString()))
                 .collect(Collectors.toList());
-        return files;
+    }
+
+    public AppServiceFile getFileByPath(String path) {
+        final File file = new File(path);
```
andxu, line 112: for get a file, why here we list all the files in the parent folder, and then get the target from a large number of files?

Flanker32: If we call `api/vfs/{path}/` to file directly, it will download the file not get the file meta data

andxu (approved):

PR #1489, opened 2021-05-10
[ASC Track2 #2] spring cloud maven/lib telemetry

wangmingliang-ms:


azure-toolkit-libs/azure-toolkit-springcloud-lib/src/main/java/com/microsoft/azure/toolkit/lib/springcloud/task/DeploySpringCloudAppTask.java
```diff
@@ -69,6 +70,7 @@ public DeploySpringCloudAppTask(SpringCloudAppConfig appConfig) {
         final boolean toCreateApp = !app.exists();
         final boolean toCreateDeployment = !deployment.exists();
 
+        AzureTelemetry.getContext().getActionProperties().put("subscriptionId", config.getSubscriptionId());
```
Flanker32, line 73: How about track subscription id while get azure client in case subscription is not set in `pom.xml` using the default value.

andxu (approved):

PR #1494, opened 2021-05-12
share resource manager factory.

shenqianjin:


azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/AzureResourceManagerFactory.java
```diff
@@ -0,0 +1,44 @@
...
+import com.microsoft.azure.toolkit.lib.Azure;
+import com.microsoft.azure.toolkit.lib.AzureConfiguration;
+import com.microsoft.azure.toolkit.lib.common.cache.Cacheable;
+
+import java.util.Optional;
+
+public class AzureResourceManagerFactory {
+
+    @Cacheable(cacheName = "AzureResourceManager", key = "$subscriptionId")
+    public static AzureResourceManager produce(String subscriptionId) {
```
andxu, line 21: produce -> create

PR #1512, opened 2021-05-14
[Track2 Function#3] Unify interfaces in app service library

Flanker32:
- Migrate to use IAzureEntityManager instead of IResource
- Move creator/updater interface to correspond resource interface


azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/service/impl/AbstractAppService.java
```diff
@@ -43,6 +57,21 @@ public void restart() {
         getRemoteResource().restart();
     }
 
+    @Override
+    public String name() {
+        return getRemoteResource().name();
```
wangmingliang-ms, line 62: IMO, we should also allow get name of a local entity
(which has no corresponding remote resource yet)

Flanker32: Using remote here in case entity is not refreshed. Have fixed the entity refresh issue in later [PR](`https://github.com/microsoft/azure-maven-plugins/pull/1521/files?file-filters%5B%5D=.java#diff-22ba8e54274abe96ccdd166d57a2f0dd5d2889d95cef4017bf55b3d6cbac1869`), will remove the unnecessary overrides later

wangmingliang-ms (approved):

PR #1516, opened 2021-05-17
[Track2 Function#7] Support create/update function deployment slot in app service library

Flanker32:
Support create/update function deployment slot in app service library


shenqianjin (approved):
As reviewed in the meeting --> It's ok to merge.

PR #1540, opened 2021-05-24
can not create new app if scaling setting is configured.

wangmingliang-ms:


azure-toolkit-libs/azure-toolkit-springcloud-lib/src/main/java/com/microsoft/azure/toolkit/lib/springcloud/SpringCloudDeployment.java
```diff
@@ -208,29 +225,13 @@ public Updater(SpringCloudDeployment deployment) {
         @AzureOperation(name = "springcloud|deployment.update", params = {"this.deployment.name()", "this.deployment.app.name()"}, type = AzureOperation.Type.SERVICE)
         public SpringCloudDeployment commit() {
             final IAzureMessager messager = AzureMessager.getMessager();
-            if (this.skippable) {
-                messager.info(String.format("Skip updating deployment(%s) since its properties is not changed.", this.deployment.name()));
-            } else {
+            if (!this.skippable) {
```
andxu, line 228: will not print "Skip updating deployment(%s) since its properties is not changed." when the scale doesn't change

wangmingliang-ms: reverted.

andxu (approved):

PR #1541, opened 2021-05-24
Fix infinite recursion issue while get slot from entity

Flanker32:
Fix infinite recursion issue while get slot from entity


azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/service/impl/FunctionAppDeploymentSlot.java
```diff
@@ -75,7 +75,7 @@ protected FunctionDeploymentSlot remote() {
 
     private FunctionApp getParentFunctionApp() {
         return StringUtils.isNotEmpty(entity.getId()) ?
-                azureClient.functionApps().getById(ResourceId.fromString(entity().getId()).parent().id()) :
+                azureClient.functionApps().getById(ResourceId.fromString(entity.getId()).parent().id()) :
```
wangmingliang-ms, line 78: if `functionApp` is passed as constructor parameter, this remote call can be saved. right?

wangmingliang-ms (approved):

PR #1548, opened 2021-05-26
bump version for bug fix.

shenqianjin:


pom.xml
```diff
@@ -67,7 +67,7 @@
         <maven.shared-utils.version>3.2.1</maven.shared-utils.version>
         <maven.version>3.3.3</maven.version>
 
-        <azure.toolkit-lib.version>0.7.0</azure.toolkit-lib.version>
+        <azure.toolkit-lib.version>0.7.1</azure.toolkit-lib.version>
```
wangmingliang-ms, line 70: this is not need. as maven plugin doesn't need this fix

shenqianjin: as discussed in the office with Andy and Hanxiao, upgrade them in a same version.
we can re-consider it again tomorrow if you have any concern.

andxu (approved):

pom.xml
andxu: Remove change of this pom file.

andxu (approved):

PR #1549, opened 2021-05-26
Skip deployment for docker function

Flanker32:
Skip deployment for docker function, in case docker deploy will broken if no artifact founded in target(which is not required for docker)


azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -379,6 +380,10 @@ protected IFunctionAppDeploymentSlot updateDeploymentSlot(final IFunctionAppDepl
     }
 
     private void deployArtifact(IFunctionAppBase target) throws AzureExecutionException {
+        if (target.getRuntime().getOperatingSystem() == OperatingSystem.DOCKER) {
```
andxu, line 383: shall we check target.getRuntime()  == null?

Flanker32: target.getRuntime() should not be null, I may add nonnull annotation later

shenqianjin (approved):

PR #1550, opened 2021-05-26
Show warning instead of exception when list triggers failed

Flanker32:
Show warning instead of exception when list triggers failed


azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -136,6 +136,7 @@
     private static final String INVALID_OS = "The value of <os> is not correct, supported values are: windows, linux and docker.";
     private static final String INVALID_JAVA_VERSION = "Unsupported value %s for <javaVersion> in pom.xml";
     private static final String INVALID_PRICING_TIER = "Unsupported value %s for <pricingTier> in pom.xml";
+    private static final String FAILED_TO_LIST_TRIGGERS = "Deployment succeeded, but failed to list http trigger urls.";
```
andxu, line 139: Deploy succeeded

Flanker32: Refers https://github.com/microsoft/vscode-azurefunctions/blob/main/src/commands/deploy/notifyDeployComplete.ts#L53

andxu (approved):

PR #1553, opened 2021-05-31
[Task 1848099]add to string for common models and add list webapp run…

andxu:
…time functionality

### DIFF
--- a/azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/AzureAppService.java
+++ b/azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/AzureAppService.java
@@ -16,6 +16,9 @@ import com.microsoft.azure.toolkit.lib.appservice.entity.AppServicePlanEntity;
 import com.microsoft.azure.toolkit.lib.appservice.entity.FunctionAppEntity;
 import com.microsoft.azure.toolkit.lib.appservice.entity.WebAppDeploymentSlotEntity;
 import com.microsoft.azure.toolkit.lib.appservice.entity.WebAppEntity;
+import com.microsoft.azure.toolkit.lib.appservice.model.JavaVersion;
+import com.microsoft.azure.toolkit.lib.appservice.model.OperatingSystem;
+import com.microsoft.azure.toolkit.lib.appservice.model.Runtime;
 import com.microsoft.azure.toolkit.lib.appservice.service.IAppServicePlan;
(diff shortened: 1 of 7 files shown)


azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/AzureAppService.java
```diff
@@ -104,6 +108,11 @@ public IWebApp webapp(WebAppEntity webAppEntity) {
                 .collect(Collectors.toList());
     }
 
+    public @Nonnull List<Runtime> listWebAppRuntimes(@Nonnull OperatingSystem os, @Nonnull JavaVersion version) {
+        return Runtime.values().stream().filter(r -> Objects.equals(os, r.getOperatingSystem()) && Objects.equals(version, r.getJavaVersion()))
```
Flanker32, line 112: `Runtime` includes runtime for web app and functions, we may need to add correspond constant/static method for each service

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/model/WebContainer.java
```diff
@@ -82,4 +83,16 @@ public boolean equals(Object target) {
     public int hashCode() {
         return Objects.hash(value);
     }
+
+    @Override
+    public String toString() {
+        if (this.equals(JAVA_OFF)) {
```
Flanker32, line 89: do we need to handle `JAVA_OFF` here? If so, do we also need to handle `JAVA_OFF` in `JavaVersion`

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/AzureAppService.java
andxu: this PR is only for webapp

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/model/WebContainer.java
Flanker32: May modify value in constructor and return value directly in `toString`

Flanker32 (approved):
todo:

Refactor enum with name/label interface

Flanker32 (dismissed):

Flanker32 (approved):

PR #1558, opened 2021-06-01
[Task 1847345] Unique the version of same dependency.

shenqianjin:
[AB#1847345](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1847345)


azure-maven-plugin-lib/pom.xml
```diff
@@ -142,6 +142,22 @@
             <groupId>org.dom4j</groupId>
             <artifactId>dom4j</artifactId>
         </dependency>
+        <dependency>
```
Flanker32, line 145: Why we need to declare dependency for `asm` here?

shenqianjin: updated.

andxu (approved):

andxu (approved):

PR #1572, opened 2021-06-09
[#1850514] refactor AzureMessager to allow setting context properties

wangmingliang-ms:
[AB#1850514](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1850514)


andxu (approved):

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/task/AzureTaskContext.java
```diff
@@ -89,7 +91,11 @@ public static void run(final Runnable runnable, AzureTaskContext context) {
                 assert Objects.equals(task, popped) : String.format("popped op[%s] is not the exiting async task[%s]", popped, task);
             });
         } catch (final Throwable throwable) {
-            AzureExceptionHandler.onRxException(throwable);
+            final Throwable rootCause = ExceptionUtils.getRootCause(throwable);
+            if (!(rootCause instanceof InterruptedIOException) &&!(rootCause instanceof InterruptedException)) {
```
Flanker32, line 95: Could we swallow `InterruptedException` in all cases? Or add an exception filter in messager context (May default to filter out interrupt exceptions

Flanker32 (approved):

PR #1589, opened 2021-06-18
[#1847674] enable re-preloading whenever user sign-in and re-selecte subscriptions

wangmingliang-ms:
[AB#1847674](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1847674)

### DIFF
--- a/azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/cache/Preloader.java
+++ b/azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/cache/Preloader.java
@@ -5,6 +5,7 @@
 
 package com.microsoft.azure.toolkit.lib.common.cache;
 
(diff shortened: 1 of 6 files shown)


azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/cache/Preloader.java
```diff
@@ -64,7 +64,8 @@ private static void invoke(final Method m, final Object instance) {
     private static Object getSingleton(final Method m) {
         final Class<?> clazz = m.getDeclaringClass();
         try {
-            final Method getInstance = clazz.getDeclaredMethod("getInstance");
+            final String methodName = AzureService.class.isAssignableFrom(clazz) ? "az" : "getInstance";
```
Flanker32, line 67: May call `AzureService` with `Azure.az(AzureService.class)...` here directly

PR #1621, opened 2021-07-06
[Task 1857855] Validate configuration in `webapp:config` and show warning

Flanker32:
Validate configuration in `webapp:config` and show warning


andxu:
add task id

andxu (approved):

PR #1658, opened 2021-07-21
[Bug fix] handle null values for creating a new service plan

andxu:


azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/task/CreateOrUpdateAppServicePlanTask.java
```diff
@@ -28,12 +29,23 @@
...
+                throw new AzureToolkitRuntimeException("Missing required configuration for 'runtime.os'.");
+            }
+
+            if (config.region() == null) {
+                throw new AzureToolkitRuntimeException("Missing required configuration for 'region'.");
+            }
+
+            if (config.pricingTier() == null) {
+                throw new AzureToolkitRuntimeException("Missing required configuration for 'pricingTier'.");
+            }
```
shenqianjin, line 47: why do these case happen ?
I think we'd better to find out why they come out and prevent that (or provide default values) rather than show a error to user.

andxu: It will be handled in future feature of advanced valuation, at here, now add the null check to avoid show npe to user.

shenqianjin (approved):
LGTM

PR #1661, opened 2021-07-22
remove the `because` before exception message

wangmingliang-ms:


azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/messager/AzureMessage.java
```diff
@@ -66,9 +65,9 @@ public String getContent() {
         final String failure = operations.stream().findFirst().map(IAzureOperation::getTitle)
                 .map(azureString -> "Failed to " + this.decorateText(azureString, azureString::getString)).orElse("Failed to proceed");
         final String cause = Optional.ofNullable(this.getCause(throwable))
-                .map(c -> String.format(", because %s", c))
+                .map(c -> "," + c + ".")
```
Flanker32, line 68: add `.` only if it was missed in cause?

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/messager/AzureMessage.java
```diff
@@ -66,9 +65,9 @@ public String getContent() {
...
                 .orElse("");
-        final String errorAction = Optional.ofNullable(this.getErrorAction(throwable)).orElse("");
+        final String errorAction = Optional.ofNullable(this.getErrorAction(throwable)).map(a -> System.lineSeparator() + a).orElse("");
```
Flanker32, line 70: As error action is in new line, do we need to call `capitalize` here?

Flanker32 (approved):

PR #1663, opened 2021-07-23
Fix regex pattern for web app validation

Flanker32:
Fix regex pattern for web app validation


andxu (approved):
Be careful next time and validate through https://regex101.com/

PR #1664, opened 2021-07-23
picture will not be loaded because of wrong base url.

wangmingliang-ms:
MicrosoftTeams-image (1)
MicrosoftTeams-image (2)


andxu (approved):
LGTM, is it a change from server side?

PR #1669, opened 2021-07-26
Refactor method for list functions and force refresh during retry

Flanker32:
Refactor method for list functions and force refresh during retry

### DIFF
--- a/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
+++ b/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
@@ -28,6 +28,7 @@ import com.microsoft.azure.toolkit.lib.appservice.service.IFunctionApp;
 import com.microsoft.azure.toolkit.lib.appservice.service.IFunctionAppBase;
 import com.microsoft.azure.toolkit.lib.appservice.service.IFunctionAppDeploymentSlot;
 import com.microsoft.azure.toolkit.lib.auth.AzureAccount;
+import com.microsoft.azure.toolkit.lib.common.bundle.AzureString;
 import com.microsoft.azure.toolkit.lib.common.exception.AzureExecutionException;
 import com.microsoft.azure.toolkit.lib.common.exception.AzureToolkitRuntimeException;
 import com.microsoft.azure.toolkit.lib.common.messager.AzureMessager;
@@ -43,13 +44,18 @@ import org.apache.maven.artifact.versioning.ComparableVersion;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.zeroturnaround.zip.ZipUtil;
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
+import reactor.util.retry.Retry;
 
 import java.io.File;
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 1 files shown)


andxu (dismissed):

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -113,9 +119,10 @@
     private static final String UPDATE_FUNCTION_DONE = "Successfully updated Function App %s.";
     private static final String NO_ARTIFACT_FOUNDED = "Failed to find function artifact '%s.jar' in folder '%s', please re-package the project and try again.";
     private static final String LOCAL_SETTINGS_FILE = "local.settings.json";
-    private static final int LIST_TRIGGERS_MAX_RETRY = 3;
+    private static final int LIST_TRIGGERS_MAX_RETRY = 4;
```
andxu, line 122: odd number 4

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -440,24 +447,16 @@ protected void listHTTPTriggerUrls(IFunctionApp target) {
...
-                // swallow service exception while list triggers
-            }
-            try {
-                Thread.sleep(LIST_TRIGGERS_RETRY_PERIOD_IN_SECONDS * 1000);
-            } catch (InterruptedException e) {
-                // swallow interrupted exception
-            }
-        }
-        throw new AzureToolkitRuntimeException(NO_TRIGGERS_FOUNDED);
+        final AtomicInteger count = new AtomicInteger(0);
```
wangmingliang-ms, line 450: use `AtomicBoolean` if needed, and maybe `boolean` is enough

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
```diff
@@ -440,24 +447,16 @@ protected void listHTTPTriggerUrls(IFunctionApp target) {
...
+        return Mono.fromCallable(() -> {
+            final AzureString message = count.getAndAdd(1) == 0 ?
+                    AzureString.fromString(SYNCING_TRIGGERS) : AzureString.format(SYNCING_TRIGGERS_WITH_RETRY, count.get(), LIST_TRIGGERS_MAX_RETRY);
+            AzureMessager.getMessager().info(message);
+            return Optional.ofNullable(functionApp.listFunctions(true))
+                    .filter(CollectionUtils::isNotEmpty)
+                    .orElseThrow(() -> new AzureToolkitRuntimeException(NO_TRIGGERS_FOUNDED));
+        }).subscribeOn(Schedulers.boundedElastic())
+                .retryWhen(Retry.backoff(LIST_TRIGGERS_MAX_RETRY - 1, Duration.ofSeconds(LIST_TRIGGERS_RETRY_PERIOD_IN_SECONDS))).block();
```
wangmingliang-ms, line 459: will `block()` throw exception `NO_TRIGGERS_FOUNDED` if the last try still fails?

Flanker32: Will throw `reactor.core.Exceptions$RetryExhaustedException` while no triggers found was the cause. And this exception will be caught by called and show warning for users

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
Flanker32: Need a final indicator in lambda, so use `AtomicInteger` here

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/DeployMojo.java
Flanker32: Refers https://github.com/microsoft/vscode-azurefunctions/blob/main/src/commands/deploy/notifyDeployComplete.ts#L40, 1 normal call + 3 retry

andxu (approved):

PR #1671, opened 2021-07-26
Set resource as un-exist only if got 404 exception for get request

Flanker32:
Set resource as un-exist only if got 404 exception for get request


andxu (approved):
LGTM

PR #1673, opened 2021-07-29
Fix bug: print '!!' for empty string in messager 

andxu:
## What problem is this PR solving?
print '!!' when calling `messager.info("")`

## What has been done in this PR?
 1. return empty string for empty string before using name to get pattern
 2. use a unified MIT header


azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/bundle/AzureString.java
```diff
@@ -47,6 +46,11 @@ public String getString() {
     }
 
     public String getString(Object... params) {
+        // sometimes we may need to print an empty line by: messager.info("")
+        // we need to
+        if (StringUtils.isBlank(name)) {
+            return "";
+        }
```
wangmingliang-ms, line 53: This is not the "right" way to print a empty line.
And I think this issue is not important/severe enough to be fixed as a hotfix

PR #1677, opened 2021-08-05
[Task 1862512] support username and password in proxy

andxu:
## What problem is this PR solving?
Current proxy doesn't support username and password, this blocks some proxy users with username and password.

## What has been done in this PR?
 1.  add username and password in az config
 2.  read username and password from maven proxy settings
 3.  apply username and password in Configuration.PROPERTY_HTTP_PROXY
 4.  apply the proxy from -D parameters for maven plugins: https://stackoverflow.com/questions/1626549/authenticated-http-proxy-with-java

The username password part has been updated in [wiki](https://github.com/microsoft/azure-maven-plugins/wiki/Proxy)

[AB#1862512](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1862512)


azure-maven-plugin-lib/src/main/java/com/microsoft/azure/maven/AbstractAzureMojo.java
```diff
@@ -517,6 +519,26 @@ public void execute() throws MojoExecutionException {
...
+                    proxy -> proxy.isActive() && proxy.getPort() > 0 && StringUtils.isNotBlank(proxy.getHost())).findFirst().orElse(null);
+                if (mavenProxy != null) {
+                    final ProxyManager.ProxyInfo mavenProxyInfo = ProxyManager.ProxyInfo.builder()
+                        .source("maven")
+                        .host(mavenProxy.getHost())
+                        .port(mavenProxy.getPort())
+                        .username(mavenProxy.getUsername())
+                        .password(mavenProxy.getPassword())
+                        .build();
+                    ProxyManager.getInstance().setActiveProxy(mavenProxyInfo);
```
Flanker32, line 536: It seems that we need to call `ProxyManager.getInstance().applyProxy();` after set Active proxy, so shall we invoke apply method in `setActiveProxy`?

Flanker32: Or shall we set active proxy to `AzureConfiguration` directly and then call `ProxyManager.getInstance().applyProxy();`

As its a bit confused that we set proxy to `ProxyManager` but get proxy configuration from Azure Configuration in `TokenCredentialManager` and `AzureCliUtils`

andxu: updated

Flanker32 (approved):

PR #1684, opened 2021-08-09
add interfaces HasUrl/Removable/Startable for action system

shenqianjin:


azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/entity/HasUrl.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ */
+
+package com.microsoft.azure.toolkit.lib.common.entity;
+
+public interface HasUrl {
```
andxu, line 8: delete this interface since it is not common

PR #1693, opened 2021-08-10
[Bug fix] fix wrong value for `authLevel` in function schema

Flanker32:
### Issue to Resolve
Values for `authLevel` is in lower cases in schema, differs with values in [AuthorizationLevel](https://github.com/Azure/azure-functions-java-library/blob/dev/src/main/java/com/microsoft/azure/functions/annotation/AuthorizationLevel.java), which may have compile issue if we use values in schema directly.

### Solution
Keep values in schema consistent with values in correspond enum


andxu (approved):
LGTM

PR #1699, opened 2021-08-16
[Task 1862943] new database-lib to extract common codes between MySQL and SQL

shenqianjin:
[Task 1862943] new database-lib to extract common codes between MySQL and SQL:

1. new database-lib module.
2. move database files under common-lib database folder to database lib.
3. remove ISqlServer interface.


andxu (approved):
LGTM

PR #1704, opened 2021-08-17
[Task 1868489] refactor mysql-lib to keep unified module style on it

shenqianjin:
[AB#1868489](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1868489)


azure-toolkit-libs/azure-toolkit-mysql-lib/src/main/java/com/microsoft/azure/toolkit/lib/mysql/model/MySqlServerConfig.java
```diff
@@ -0,0 +1,44 @@
...
+import com.microsoft.azure.toolkit.lib.database.entity.IDatabaseServerEntity;
+import lombok.Builder;
+import lombok.Getter;
+
+import java.util.Objects;
+
+@Getter
+@Builder
+public class MySqlServerConfig implements IDatabaseServerEntity {
+
```
andxu, line 20: if it is the same with SqlServerConfig, can we use the same class?

shenqianjin: yes. as files of different resources domain, I prefer to keep them as different classes

andxu (approved):
LGTM

PR #1706, opened 2021-08-17
Fix validation issue for service principal authentication with `<serverId>`

Flanker32:
### Issues to resolve
We use to share schema between `AuthConfiguration` and `MavenAuthConfiguration`, however, `MavenAuthConfiguration` support get client/tenant/key from `<servers>` in maven `settings.xml`, which was not supported in `AuthConfiguration`, and users will meet following exception when they auth maven plugin with service principal in settings.xml

image

### Solution
Create new schema for maven plugin auth configuration, support  `<serverId>` for service principal authentication

### DIFF
--- a/azure-maven-plugin-lib/src/main/resources/schema/maven/AzureAppServiceMavenPlugin.json
+++ b/azure-maven-plugin-lib/src/main/resources/schema/maven/AzureAppServiceMavenPlugin.json
@@ -63,16 +63,88 @@
     },
     "auth": {
       "title": "AuthConfiguration",
-      "description": "The auth config for accessing azure resources for maven",
+      "$schema": "http://json-schema.org/draft-07/schema#",
+      "description": "The auth config for accessing azure resources",
       "type": "object",
       "properties": {
+        "type": {
+          "$ref": "classpath:///schema/common/AuthConfiguration.json#/definitions/auth-type"
+        },
+        "client": {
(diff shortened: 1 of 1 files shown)


azure-maven-plugin-lib/src/main/resources/schema/maven/AzureAppServiceMavenPlugin.json
```diff
@@ -63,16 +63,88 @@
...
+      "$schema": "http://json-schema.org/draft-07/schema#",
+      "description": "The auth config for accessing azure resources",
       "type": "object",
       "properties": {
+        "type": {
+          "$ref": "classpath:///schema/common/AuthConfiguration.json#/definitions/auth-type"
+        },
+        "client": {
+          "description": "Client ID",
+          "type": "string"
```
andxu, line 75: guid

azure-maven-plugin-lib/src/main/resources/schema/maven/AzureAppServiceMavenPlugin.json
```diff
@@ -63,16 +63,88 @@
...
+        },
+        "tenant": {
+          "description": "Tenant ID",
+          "type": "string"
```
andxu, line 79: guid regex

andxu (approved):
LGTM

PR #1711, opened 2021-08-18
Validate `CreateOrUpdateAppServicePlanTask` with SchemaValidator

Flanker32:
### Issues to resolve
Our current schema validation for web app command line plugin will only affect on the startup, so we could not determine whether to create or update existing resource, and could not apply validation for resource creation. 

### Solution
Create multi schema for different tasks, we could reuse existing schema with `ref` and "extend" existing schema with `allOf`

### Commits
- Support validate and throw exception in `SchemaValidator`
- Add new schema for app service plan creation/update and validate configuration during task execution
- Add schema relative path to id and fix reflection issues when no schema founded

### DIFF
--- a/azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/schema/appservice/AppServicePlan.json
+++ b/azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/schema/appservice/AppServicePlan.json
@@ -1,0 +1,33 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "App Service Plan - Update",
+  "description": "Schema for AppServicePlanConfig when update existing service plan",
+  "properties": {
+    "subscriptionId": {
+      "$ref": "classpath:///schema/common/UUID.json"
+    },
+    "servicePlanResourceGroup": {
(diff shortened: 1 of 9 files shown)


azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/schema/appservice/AppServicePlan.json
```diff
@@ -0,0 +1,33 @@
...
+      "$ref": "classpath:///schema/appservice/AppServicePlanName.json"
+    },
+    "region": {
+      "$ref": "classpath:///schema/common/NonEmptyString.json"
+    },
+    "os": {
+      "$ref": "classpath:///schema/appservice/Runtime.json#/definitions/os"
+    },
+    "region": {
+      "type": "object"
```
andxu, line 22: why two region?

andxu: type object?

Flanker32: Region contains two properties, value and display name, so it will convert to json object when serialize. And the duplicated property has been removed

PR #1719, opened 2021-08-23
Revert "unified azure operation title properties key style."

shenqianjin:
Reverts microsoft/azure-maven-plugins#1718


andxu (approved):
LGTM

PR #1722, opened 2021-08-24
Fix the identity proxy caused by ProxyUtil refactor

andxu:


azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/util/AzureEnvironmentUtils.java
```diff
@@ -52,6 +52,17 @@ public static void setupAzureEnvironment(AzureEnvironment env) {
             // src/main/java/com/azure/identity/implementation/IdentityClientOptions.java#L42
             Configuration.getGlobalConfiguration().put(Configuration.PROPERTY_AZURE_AUTHORITY_HOST, env.getActiveDirectoryEndpoint());
         }
+
+        final com.microsoft.azure.toolkit.lib.AzureConfiguration az = com.microsoft.azure.toolkit.lib.Azure.az().config();
```
Flanker32, line 56: Please make sure the method is invoked in each tooling

Flanker32 (approved):

Flanker32 (approved):

azure-toolkit-libs/azure-toolkit-auth-lib/src/main/java/com/microsoft/azure/toolkit/lib/auth/util/AzureEnvironmentUtils.java
andxu: Auth lib will call this method before every login

PR #1729, opened 2021-08-30
Migrate to AppServiceManager in app service library

Flanker32:
### Issues to resolve
Now we leverage `AzureResourceManager` directly in app service library, which will initialize many modules which was not used in library and may slow down the performance of Eclipse plugin
image

### Solution
Use `AppServiceManager` instead of `AzureResourceManager`

### Next Steps
investigate why Eclipse has much worse performance than IntelliJ when initialize `AzureResourceManager`

### DIFF
--- a/azure-toolkit-libs/azure-toolkit-applicationinsights-lib/src/main/java/com/microsoft/azure/toolkit/lib/applicationinsights/ApplicationInsights.java
+++ b/azure-toolkit-libs/azure-toolkit-applicationinsights-lib/src/main/java/com/microsoft/azure/toolkit/lib/applicationinsights/ApplicationInsights.java
@@ -94,7 +94,7 @@ public class ApplicationInsights extends SubscriptionScoped<ApplicationInsights>
         getApplicationInsightsManager(subscriptionId).components().deleteByResourceGroup(resourceGroup, name);
     }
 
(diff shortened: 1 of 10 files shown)


azure-toolkit-libs/azure-toolkit-applicationinsights-lib/src/main/java/com/microsoft/azure/toolkit/lib/applicationinsights/ApplicationInsights.java
```diff
@@ -94,7 +98,7 @@ public void delete(@Nonnull String subscriptionId, @Nonnull String resourceGroup
         getApplicationInsightsManager(subscriptionId).components().deleteByResourceGroup(resourceGroup, name);
     }
 
-    @Cacheable(cacheName = "ApplicationInsightsManager", key = "$subscriptionId")
+    @Cacheable(cacheName = "AzureResourceManager", key = "$subscriptionId")
```
andxu, line 101: wrong name?

Flanker32: Fixed

andxu (approved):
LGTM

PR #1733, opened 2021-09-03
Update rules for check style

Flanker32:
### Issues to resolve
We have updated the check style CI pipeline, but rules was not updated in maven configuration, which may break the CI build even if we pass the check style ci.

### Solution
Update the check style rules to keep consistent with CI, refers https://github.com/andxu/java-checkstyle/tree/master/checkstyle


andxu (approved):
LGTM

PR #1767, opened 2021-09-23
Support storage and Azure spot in compute library

Flanker32:
- Support storage and Azure spot during VM creation
- Add no args constructor for draft resources
- Add constants for parameters and default constructor for draft resources
- Resolve issues in tests


andxu (approved):
LGTM

PR #1773, opened 2021-09-23
remove track1 sdk dependency

andxu:


Flanker32 (approved):
Please add the dependencies in Utils

PR #1775, opened 2021-09-24
[#1880367] popup login dialog when user execute an action requiring auth but not authenticated.

wangmingliang-ms:
[AB#1880367](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1880367)


andxu (approved):
LGTM

PR #1778, opened 2021-09-26
Fix a bug of cannot update AI key

andxu:


andxu:
replaced with https://github.com/microsoft/azure-maven-plugins/pull/1779

PR #1779, opened 2021-09-26
Fix function maven plugin update issues

Flanker32:
- Fix app name was used for deployment slot by mistake
- Fix users ai key was replaced by default value


andxu (approved):
LGTM

PR #1780, opened 2021-09-26
Fix telemetry properties was missed in app service maven tooling

Flanker32:
Fix telemetry properties was missed in app service maven tooling


wangmingliang-ms (approved):

andxu (approved):
LGTM

PR #1786, opened 2021-09-29
Fix null version issue in sql server entity.

andxu:


andxu:
the change in https://github.com/microsoft/azure-maven-plugins/pull/1781 shall be released in september release

Flanker32 (approved):
LGTM

PR #1792, opened 2021-10-11
Refresh service manager after resource creation

Flanker32:
Refresh service manager after resource creation in app service library


Flanker32:
move to #1793

PR #1794, opened 2021-10-13
#1885355: key of azure storage account is not correct in resource con…

wangmingliang-ms:
[AB#1885355](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1885355)


andxu (approved):
LGTM

PR #1795, opened 2021-10-15
Skip extension install with extension bundle preview

### LINKED ISSUE #112: azure-functions-gradle-plugin should not try to download ExtensionBundle.Preview
I'm working on Linux/Java8 setup with Azure Functions.

In order to use UAMI with Blob Trigger, I'm testing the latest ExtensionBundle.Preview as discussed on [#6423](https://github.com/Azure/azure-functions-host/issues/6423#issuecomment-876693586) (it is under Azure organization).

When building the with azureFunctionsPackageZip task, the plugin is trying to install the extension with functions core tools. Per the current login in [PackageHandler](https://github.com/microsoft/azure-gradle-plugins/blob/master/azure-functions-gradle-plugin/src/main/java/com/microsoft/azure/plugin/functions/gradle/handler/PackageHandler.java), it will not download with func core tools if host.json is declared with ExtensionBundle. IMO, the same logic should apply for ExtensionBundle.Preview as well.

Can we align the handling between ExtensionBundle and ExtensionBundle.Preview?

Flanker32:
Fixes https://github.com/microsoft/azure-gradle-plugins/issues/112

[AB#1885843](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1885843)


andxu (approved):
LGTM

PR #1811, opened 2021-11-03
#1890610: add `Add SSH Configuration` action and open SSH configuration dialog on act

wangmingliang-ms:
[AB#1890610](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1890610): add `Add SSH Configuration` action and open SSH configuration dialog on act

### DIFF
--- a/azure-toolkit-libs/azure-toolkit-compute-lib/src/main/java/com/microsoft/azure/toolkit/lib/compute/vm/VirtualMachine.java
+++ b/azure-toolkit-libs/azure-toolkit-compute-lib/src/main/java/com/microsoft/azure/toolkit/lib/compute/vm/VirtualMachine.java
@@ -5,6 +5,7 @@
 
 package com.microsoft.azure.toolkit.lib.compute.vm;
 
+import com.azure.resourcemanager.compute.models.OSProfile;
 import com.azure.resourcemanager.compute.models.PowerState;
 import com.microsoft.azure.toolkit.lib.common.entity.IAzureBaseResource;
 import com.microsoft.azure.toolkit.lib.common.entity.IAzureModule;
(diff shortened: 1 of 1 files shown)


azure-toolkit-libs/azure-toolkit-compute-lib/src/main/java/com/microsoft/azure/toolkit/lib/compute/vm/VirtualMachine.java
```diff
@@ -65,6 +66,36 @@ public Region getRegion() {
...
+        // return remote().getPrimaryNetworkInterface().getNetworkSecurityGroup().securityRules().entrySet().stream()
+        //    .anyMatch(e -> SecurityRuleProtocol.TCP.equals(e.getValue().protocol()) &&
+        //        SecurityRuleDirection.INBOUND.equals(e.getValue().direction()) &&
+        //        "22".equals(e.getValue().destinationPortRange()) &&
+        //        "*".equals(e.getValue().destinationPortRange()));
+        return true;
+    }
+
+    public String getHostIp() {
+        return remote().getPrimaryPublicIPAddress().ipAddress();
```
Flanker32: VM may not have a public ip address, we may need to add null check here. Besides, do we also need to add method to create public ip address?

azure-toolkit-libs/azure-toolkit-compute-lib/src/main/java/com/microsoft/azure/toolkit/lib/compute/vm/VirtualMachine.java
```diff
@@ -65,6 +66,36 @@ public Region getRegion() {
...
+    }
+
+    public String getAdminUserName() {
+        return remote().innerModel().osProfile().adminUsername();
+    }
+
+    public String getAdminPassword() {
+        return remote().innerModel().osProfile().adminPassword();
```
Flanker32: Could we get password from arm api? The result seems to be null.

azure-toolkit-libs/azure-toolkit-compute-lib/src/main/java/com/microsoft/azure/toolkit/lib/compute/vm/VirtualMachine.java
```diff
@@ -65,6 +66,36 @@ public Region getRegion() {
...
+    }
+
+    public boolean isPasswordAuthenticationDisabled() {
+        final OSProfile profile = remote().innerModel().osProfile();
+        if (Objects.nonNull(profile.linuxConfiguration())) {
+            return profile.linuxConfiguration().disablePasswordAuthentication();
+        }
+        return true;
```
Flanker32: return false here? As password should be the only authentication method for windows vm

Flanker32 (changes requested):

Flanker32 (approved):
LGTM

PR #1833, opened 2021-11-19
Migrate to use ConsoleTextTerminal in web app config mojo

Flanker32:
Migrate to use ConsoleTextTerminal in web app config mojo, refers https://github.com/microsoft/azure-maven-plugins/pull/1830


andxu (approved):
LGTM

PR #1834, opened 2021-11-20
#1896362: [Test]When click More Settings when creating a Function App, the OK button does not work.

wangmingliang-ms:
[AB#1896362](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1896362): [Test]When click More Settings when creating a Function App, the OK button does not work.


Flanker32 (approved):
LGTM

PR #1840, opened 2021-11-24
Fix authentication issue in cloud shell

Flanker32:
- Bump Azure dependency version for maven toolkits
- Hot fix for azure cli authentication in cloud shell
- Fix queryer did not work in git bash


andxu (approved):
LGTM

PR #1850, opened 2021-12-05
Remove redundant interface for app service and replace with implementation

Flanker32:
Remove redundant interface for app service/service plan and replace with implementation


wangmingliang-ms (approved):
LGTM

PR #1857, opened 2021-12-10
[#1888736] Fix status sync issue for app service

Flanker32:
- Keep cache for get when refresh app service entity
- Leverage cache for get when list resources


andxu (approved):
LGTM

PR #1861, opened 2021-12-14
#1902712: add "sign in" action to all AzureAuthenticationException

wangmingliang-ms:
AB#1902712: add "sign in" action to all AzureAuthenticationException


Flanker32 (approved):
LGTM

PR #1867, opened 2021-12-16
#1903163: add `Retry from failure` action on notifications of action …

wangmingliang-ms:
AB#1903163: add `Retry from failure` action on notifications of action failures when possible


andxu (approved):
LGTM

PR #1869, opened 2021-12-16
FEATURE#1776604: add reasonable actions to notifications (especially errors), so that user knows what to do next.

wangmingliang-ms:
[AB#1776604](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1776604): add reasonable actions to notifications (especially errors), so that user knows what to do next.


andxu (approved):
LGTM

PR #1878, opened 2021-12-23
Unify telemetry properties name in different telemeters

Flanker32:
Unify telemetry properties name in different telemeters


andxu (approved):
LGTM

PR #1903, opened 2022-01-11
Update azure operation properties for app service

Flanker32:
- Add missing azure operations for deployment slot and update operation properties to keep consistent with other toolkit library
- Fix deployment slot cache was not clean up after refresh app service instance

For https://github.com/microsoft/azure-tools-for-java/pull/6313


wangmingliang-ms (approved):
LGTM

PR #1910, opened 2022-01-19
Fix miscellaneous bugs related to app service library

Flanker32:
- Catch exception for list resources and show warning during list resources in account
- Cleanup cache for get/list when refresh resource module
- Support get raw entity for app service


wangmingliang-ms (approved):
LGTM

PR #1914, opened 2022-01-21
[Task #1872026] Deprecate entity in app service, move the get method to related resource

Flanker32:
### Issue to resolve
Currently we will send unnecessary requests when get runtime configuration or app settings of app service, as we use leverage the `entity()` method to get the parameters. However, entity() will load all properties of app service and in most cases we just care about specific properties like runtime, and the unnecessary requests may affect our performance.

### Solution
Add method with cache to get properties directly in app service library 

[AB#1872026](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1872026)


wangmingliang-ms (approved):
LGTM

PR #1932, opened 2022-02-14
#1917263: [Test] No any prompts when update deployment without resource template or parameters

wangmingliang-ms:


Flanker32 (approved):
LGTM

PR #1940, opened 2022-02-21
#1919630: use slf4j and log4j v2 to log

wangmingliang-ms:
`@Log4j2` doesn't work well in maven, which cause some details not correctly logged. So, in this PR, we migrate to `@Slf4j` instead, which has better compatibility with different log implementations.

[AB#1920298](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1920298): [Test] Subscription info is not right if deploy maven spring cloud with azure_cli login
[#5039](https://github.com/microsoft/azure-tools-for-java/issues/5039) : Use SLF4J For Logging

Has this been tested?
---------------------------
- [ ] Tested


Flanker32 (approved):
LGTM

PR #1941, opened 2022-02-21
[Bugfix #1917261] Fail to update deployment after editing deployment

wangmingliang-ms:
this PR always set `template` when create/update an ARM deployment.

`template` is always required when create/update an ARM deployment, but `template` is not set if it found not changed before this PR, which cause validation exception. 

[AB#1917261](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1917261): [Test] Fail to update deployment after editing deployment

Has this been tested?
---------------------------
- [ ] Tested


Flanker32 (approved):
LGTM

PR #1945, opened 2022-02-23
#1917238: [Test] IDE always blocks when refresh groups under Resource Management

wangmingliang-ms:
The IDE blocks once user expand a node which has many children nodes.
The cause is once a child resource added to its parent, a event would be fired, the AsyncEventBus consumes too many threads/CPU to fire/handle these events.
this PR fix this problem by using debouncing.

[AB#1917238](https://dev.azure.com/mseng/a4d27ce2-a42d-4b71-8eef-78cee9a9728e/_workitems/edit/1917238): [Test] IDE always blocks when refresh groups under Resource Management

Has this been tested?
---------------------------
- [x] Tested


Flanker32 (approved):
LGTM

PR #1977, opened 2022-03-23
Fix value of functionId in functionentity

Flanker32:
Fix value of functionId in functionentity

N/A

N/A

Any other comments?
-------------------
N/A

Has this been tested?
---------------------------
- [ ] Tested

### DIFF
--- a/azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/function/FunctionApp.java
+++ b/azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/function/FunctionApp.java
@@ -64,8 +64,7 @@ public class FunctionApp extends FunctionAppBase<FunctionApp, AppServiceResource
     @Nonnull
     public List<FunctionEntity> listFunctions(boolean... force) {
         return Optional.ofNullable(this.getFullRemote()).map(r -> r.listFunctions().stream()
-                        .map(envelope ->
-                                AppServiceUtils.fromFunctionAppEnvelope(envelope, (com.azure.resourcemanager.appservice.models.FunctionApp) this.getRemote()))
+                        .map(envelope -> AppServiceUtils.fromFunctionAppEnvelope(envelope, this.getId()))
                         .filter(Objects::nonNull)
                         .collect(Collectors.toList()))
                 .orElse(Collections.emptyList());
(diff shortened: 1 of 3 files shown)


azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/utils/AppServiceUtils.java
```diff
@@ -390,7 +391,7 @@ public static FunctionEntity fromFunctionAppEnvelope(@Nonnull FunctionEnvelope f
                 .entryPoint(entryPoint)
                 .scriptFile(scriptFile)
                 .bindingList(bindingEntities)
-                .functionAppId(functionEnvelope.functionAppId())
+                .functionAppId(functionApp.id())
```
wangmingliang-ms: why not use `functionEnvelope.functionAppId()`

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/function/FunctionApp.java
```diff
@@ -64,10 +64,11 @@ public String getMasterKey() {
     @Nonnull
     public List<FunctionEntity> listFunctions(boolean... force) {
         return Optional.ofNullable(this.getFullRemote()).map(r -> r.listFunctions().stream()
-                .map(AppServiceUtils::fromFunctionAppEnvelope)
-                .filter(Objects::nonNull)
-                .collect(Collectors.toList()))
-            .orElse(Collections.emptyList());
+                        .map(envelope ->
+                                AppServiceUtils.fromFunctionAppEnvelope(envelope, (com.azure.resourcemanager.appservice.models.FunctionApp) this.getRemote()))
```
wangmingliang-ms: `this.getRemote()` may send additional request to server, use `this.getId()` since only `id` is required by this util method.

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/utils/AppServiceUtils.java
Flanker32: RestAPI will not return functionAppId in list request, which will make `functionEnvelope.functionAppId()` returns null

wangmingliang-ms (approved):
LGTM

PR #2000, opened 2022-04-18
implement Azure.getOrInitById

wangmingliang-ms:
…

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested


Flanker32 (approved):
LGTM

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/model/AbstractAzService.java
```diff
@@ -106,7 +107,32 @@ protected <E> E doGetById(@Nonnull String id) {
...
+    protected <E> E doGetOrInitById(@Nonnull String id) {
+        ResourceId resourceId = ResourceId.fromString(id);
+        final String resourceGroup = resourceId.resourceGroupName();
+        AbstractAzResource<?, ?, ?> resource = Objects.requireNonNull(this.get(resourceId.subscriptionId(), resourceGroup));
+        final LinkedList<Pair<String, String>> resourceTypeNames = new LinkedList<>();
+        while (resourceId != null) {
+            resourceTypeNames.push(Pair.of(resourceId.resourceType(), resourceId.name()));
+            resourceId = resourceId.parent();
+        }
+        for (Pair<String, String> resourceTypeName : resourceTypeNames) {
```
Flanker32, line 132: May need to handle case that module is not supported, which will return null rather than a `GenericResource`

PR #2173, opened 2022-11-24
Add more customized options for functions maven plugin

### LINKED ISSUE #2075: Support passing port for function app startup
Currently we can pass to functions core tools the port for the functions to listen on using -p flag. Can we add this functionality for the azure functions plugin as well?

### LINKED ISSUE #2144: Can not use any other host.json file than ${rootDir}/host.json
### Plugin name and version
azure-functions-maven-plugin, version 1.21.0

### Plugin configuration in your `pom.xml`
(not meaningful)

### Expected behavior
If the assembly process encounters an already present host.json file at the right place in the staging directory, it should not overwrite it with its own static version.

This would allow for easier code-reuse (sharing of host.json file in multi modules, etc...), and integration with other maven  plugins (resource copying, code and resource generation, ...).

An alternative could be to have a XML config flag to prevent generation of the file and/or to specify a path for it (including build target folders).

### Actual behavior
The plugin checks the presence of the host.json in maven's project root, in which case it is copied, but in any other case, it statically overwrites it with a default template.

### Steps to reproduce the problem
1. Generate an empty function from the archetype
2. `cd` into it 
3. Add a simple host.json file, with remarkable content, e.g `mkdir static && echo '{ "version": "2.0" }' >> static/host.json`, notice how this differs from the template file
4. Alter the packaging process so as to move this static/host.json at the right place at assembly time. In the main pom, one could use the mvn resource plugin : 
```
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-resources-plugin</artifactId>
                <version>3.2.0</version>
                <executions>
                    <execution>
                        <id>copy-azure-resources</id>
                        <phase>package</phase>
                        <goals>
                            <goal>copy-resources</goal>
                        </goals>
                        <configuration>
                            <overwrite>true</overwrite>
                            <outputDirectory>${project.build.directory}/azure-functions/${functionAppName}</outputDirectory>
                            <resources>
                                <resource>
                                    <directory>${project.basedir}/static</directory>
                                    <includes>
                                        <include>host.json</include>
                                    </includes>
                                </resource>
                            </resources>
                        </configuration>
                    </execution>
                </executions>
```
5. mvn package it

One can observe that the final file is not the static/host.json one, although, if we `Ctrl-C` (or skip entirely) early before / really early in the azure plugin, the file is there.

Flanker32:
- Support customized path for `host.json`/`local.settings.json` for functions maven plugin, fixes #2144
- Support customized port for function local host, fixes #2075

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested

### DIFF
--- a/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
+++ b/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
@@ -143,14 +143,24 @@ public abstract class AbstractFunctionMojo extends AbstractAppServiceMojo {
     @Parameter(property = "functions.disableAppInsights", defaultValue = "false")
     protected boolean disableAppInsights;
 
+    /**
+     * Path for host.json file
+     *
+     * @since 1.22.0
+     */
     @Getter
     @Parameter(property = "functions.hostJson", defaultValue = HOST_JSON)
     protected String hostJson;
 
+    /**
+     * Path for local.settings.json file
+     *
+     * @since 1.22.0
+     */
     @Getter
     @Parameter(property = "functions.localSettingsJson", defaultValue = LOCAL_SETTINGS_JSON)
     protected String localSettingsJson;
-    
+
     @Getter
     protected ConfigParser parser = new ConfigParser(this);
 
--- a/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
+++ b/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
@@ -35,6 +35,7 @@ import java.util.stream.Collectors;
 public abstract class AbstractFunctionMojo extends AbstractAppServiceMojo {
 
     protected static final String HOST_JSON = "host.json";
+    protected static final String LOCAL_SETTINGS_JSON = "local.settings.json";
     protected static final String TRIGGER_TYPE = "triggerType";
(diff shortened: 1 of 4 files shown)


azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
```diff
@@ -216,7 +225,7 @@ protected String getFunctionLibraryVersion() {
 
     protected JsonNode readHostJson() {
         // todo: add configuration for host.json location
-        final File hostJson = new File(project.getBasedir(), HOST_JSON);
+        final File hostJson = new File(project.getBasedir(), getHostJson());
```
wangmingliang-ms: need to consider relative paths.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/RunMojo.java
```diff
@@ -50,6 +50,13 @@ public class RunMojo extends AbstractFunctionMojo {
     @Parameter(property = "localDebugConfig", defaultValue = "transport=dt_socket,server=y,suspend=n,address=5005")
     protected String localDebugConfig;
 
+    /**
+     * Config port for function local host
+     *
+     * @since 1.22.0
+     */
+    @Parameter(property = "funcPort", defaultValue = "7071")
+    protected Integer funcPort;
```
wangmingliang-ms, line 59: update documentation.

wangmingliang-ms (approved):
LGTM

PR #2200, opened 2023-01-03
enable cloning an Action and bind it to a specific source.

wangmingliang-ms:
…

image

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested


azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/action/Action.java
```diff
@@ -39,14 +39,12 @@
 
 @SuppressWarnings("unused")
 @Accessors(chain = true)
-public class Action<D> extends OperationBase {
+public class Action<D> extends OperationBase implements Cloneable {
```
Flanker32, line 42: Shall we use a copy constructor instead of clone() and implement Cloneable?

PR #2302, opened 2023-05-29
WIP: Generate native binary from java code

viniciusfcf:
Checkstyle is not ok.

Java projects could generate jar or native files to run (Quarkus, Micronaut or SpringBoot). Now it's very easy after accept this PR. Example with Quarkus using this plugin:  https://github.com/viniciusfcf/quarkus-azure-functions-http/tree/using-new-versions
…

No, however I'm working with Sandra Ahlgrimm to allow native function using customHandler from java projects

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested

### DIFF
--- a/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
+++ b/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
@@ -61,6 +61,7 @@ public abstract class AbstractFunctionMojo extends AbstractAppServiceMojo {
     private static final String FUNCTION_REGION_KEY = "region";
     private static final String FUNCTION_PRICING_KEY = "pricingTier";
     private static final String FUNCTION_DEPLOY_TO_SLOT_KEY = "isDeployToFunctionSlot";
+    private static final String FUNCTION_NATIVE_EXECUTABLE_PATH = "nativeExecutablePath";
 
     //region Properties
     @Parameter(defaultValue = "${project.build.finalName}", readonly = true, required = true)
@@ -214,6 +215,24 @@ public abstract class AbstractFunctionMojo extends AbstractAppServiceMojo {
     @Parameter
     protected String storageAccountResourceGroup;
 
+    /**
+     * Boolean flag to run function with custom runtime instead of java
+     *
+     * @since 1.27.0
+     */
+    @Getter
+    @Parameter(property = "functions.nativeExecutablePath")
+    protected String nativeExecutablePath;
+
+    /**
+     * Args to be used by custom handler.
+     *
+     * @since 1.27.0
+     */
+    @Getter
+    @Parameter(property = "functions.customHandlerArgs", defaultValue = "")
+    protected String customHandlerArgs;
+
     @Getter
     protected final ConfigParser parser = new ConfigParser(this);
 
@@ -269,6 +288,10 @@ public abstract class AbstractFunctionMojo extends AbstractAppServiceMojo {
             throw new AzureToolkitRuntimeException(CAN_NOT_FIND_ARTIFACT);
         }
         return result;
+    }
+
+    protected boolean isNativeExecutable() {
+        return !StringUtils.isEmpty(getNativeExecutablePath());
     }
 
     protected File getHostJsonFile() {
@@ -355,6 +378,7 @@ public abstract class AbstractFunctionMojo extends AbstractAppServiceMojo {
         result.put(DISABLE_APP_INSIGHTS_KEY, String.valueOf(isDisableAppInsights()));
         final boolean isDeployToFunctionSlot = getDeploymentSlotSetting() != null && StringUtils.isNotEmpty(getDeploymentSlotSetting().getName());
         result.put(FUNCTION_DEPLOY_TO_SLOT_KEY, String.valueOf(isDeployToFunctionSlot));
+        result.put(FUNCTION_NATIVE_EXECUTABLE_PATH, nativeExecutablePath);
         return result;
     }
     //endregion
--- a/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
+++ b/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
@@ -62,6 +62,7 @@ public abstract class AbstractFunctionMojo extends AbstractAppServiceMojo {
     private static final String FUNCTION_PRICING_KEY = "pricingTier";
     private static final String FUNCTION_DEPLOY_TO_SLOT_KEY = "isDeployToFunctionSlot";
     private static final String FUNCTION_NATIVE_EXECUTABLE_PATH = "nativeExecutablePath";
+    private static final String FUNCTION_IS_NATIVE = "isNativeFunction";
 
     //region Properties
     @Parameter(defaultValue = "${project.build.finalName}", readonly = true, required = true)
@@ -291,7 +292,7 @@ public abstract class AbstractFunctionMojo extends AbstractAppServiceMojo {
     }
 
     protected boolean isNativeExecutable() {
-        return !StringUtils.isEmpty(getNativeExecutablePath());
+        return StringUtils.isNotEmpty(getNativeExecutablePath());
     }
 
     protected File getHostJsonFile() {
@@ -378,7 +379,7 @@ public abstract class AbstractFunctionMojo extends AbstractAppServiceMojo {
         result.put(DISABLE_APP_INSIGHTS_KEY, String.valueOf(isDisableAppInsights()));
         final boolean isDeployToFunctionSlot = getDeploymentSlotSetting() != null && StringUtils.isNotEmpty(getDeploymentSlotSetting().getName());
         result.put(FUNCTION_DEPLOY_TO_SLOT_KEY, String.valueOf(isDeployToFunctionSlot));
-        result.put(FUNCTION_NATIVE_EXECUTABLE_PATH, nativeExecutablePath);
+        result.put(FUNCTION_IS_NATIVE, String.valueOf(isNativeExecutable()));
         return result;
     }
     //endregion
--- a/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
+++ b/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
@@ -29,6 +29,7 @@ import com.microsoft.azure.toolkit.lib.legacy.function.handlers.CommandHandler;
 import com.microsoft.azure.toolkit.lib.legacy.function.handlers.CommandHandlerImpl;
 import com.microsoft.azure.toolkit.lib.legacy.function.handlers.FunctionCoreToolsHandler;
 import com.microsoft.azure.toolkit.lib.legacy.function.handlers.FunctionCoreToolsHandlerImpl;
+
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections4.SetUtils;
 import org.apache.commons.io.FileUtils;
@@ -61,6 +62,7 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.nio.file.Paths;
+import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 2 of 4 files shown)


viniciusfcf:
> @microsoft-github-policy-service agree

@microsoft-github-policy-service agree company="Red Hat"

edburns:
Converting to Draft. I'll seek guidance from @Flanker32 , but in general I am in favor of this feature.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
```diff
@@ -323,6 +354,8 @@ public Map<String, String> getTelemetryProperties() {
         result.put(DISABLE_APP_INSIGHTS_KEY, String.valueOf(isDisableAppInsights()));
         final boolean isDeployToFunctionSlot = getDeploymentSlotSetting() != null && StringUtils.isNotEmpty(getDeploymentSlotSetting().getName());
         result.put(FUNCTION_DEPLOY_TO_SLOT_KEY, String.valueOf(isDeployToFunctionSlot));
+        result.put(FUNCTION_IS_NATIVE_EXECUTABLE, String.valueOf(isNativeExecutable()));
+        result.put(FUNCTION_CUSTOM_HANDLER_ARGS, customHandlerArgs);
```
Flanker32: Will ‘customHandlerArgs’ contains any PII data like file location? If so, we should not record them in telemetry but we could track whether users set custom handler args here

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
```diff
@@ -117,6 +122,10 @@ public class PackageMojo extends AbstractFunctionMojo {
     private static final String BUILD_UBER_ARTIFACT_EXCEPTION = "Failed to build uber artifact, please set `buildFatJar` to `false` and use `maven-shade-plugin` to try again.";
     //region Entry Point
 
+    public static void main(String[] args) {
```
Flanker32: Please remove test codes here

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/config/FunctionAppConfig.java
```diff
@@ -16,5 +16,7 @@ public class FunctionAppConfig extends AppServiceConfig {
     private String appInsightsInstance;
     private String appInsightsKey;
     private boolean disableAppInsights;
+    private boolean nativeExecutable;
```
Flanker32: it seems that 'nativeExecutable' and 'customHandlerArgs' are parameters used for package/run right? If so, we could remove them from 'FunctionAppConfig' as 'FunctionAppConfig' is used to create/update FunctionApp in Azure, which will not affect packaging or local run

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
```diff
@@ -228,17 +253,23 @@ protected File getArtifact() throws AzureToolkitRuntimeException {
...
-        final File result = new File(buildDirectory, StringUtils.join(finalName, FilenameUtils.EXTENSION_SEPARATOR, packaging));
+        String finalNameWithExtension = finalNameWithExtension();
+        final File result = new File(buildDirectory, finalNameWithExtension);
         if (!result.exists()) {
             throw new AzureToolkitRuntimeException(CAN_NOT_FIND_ARTIFACT);
         }
         return result;
     }
 
+    protected String finalNameWithExtension() {
```
Flanker32: `getArtifact` will be called by compile level validation and the function.json generation logic, we may be careful to modify its implementation. It seems that this method will only be called during function package, so could we just move this method to `Package.mojo`, and modify `copyBinaryToStageDirectory`to get the correct artifact?

Besides, we may add a parameter like `nativeExecutablePath' instead of boolean `nativeExecutable`, as we could only guess the final artifact here, but users may modify it with configuration in quarkus maven plugin

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
```diff
@@ -348,10 +389,15 @@ protected void copyLocalSettingsJson() throws IOException {
         log.info(SAVING_LOCAL_SETTINGS_JSON);
         final File sourceLocalSettingsJsonFile = getLocalSettingsJsonFile();
         final File destLocalSettingsJsonFile = Paths.get(getDeploymentStagingDirectoryPath(), LOCAL_SETTINGS_JSON).toFile();
-        copyFilesWithDefaultContent(sourceLocalSettingsJsonFile, destLocalSettingsJsonFile, DEFAULT_LOCAL_SETTINGS_JSON);
+        String defaultContent = MessageFormat.format(DEFAULT_LOCAL_SETTINGS_JSON, getWorkerRuntime());
```
Flanker32, line 389: Do we need to overwrite users configuration for native artifact? As default content will not work if users have a local.settings.json in their workspace. Same for the host.json

Flanker32:
@viniciusfcf Thanks a lot for your efforts! Please see my comments in this PR.

~~Besides, I tried to run your demo project https://github.com/viniciusfcf/quarkus-azure-functions-http/tree/using-new-versions, but will get same error here https://github.com/viniciusfcf/quarkus-azure-functions-http/blob/using-new-versions/image.png, could you please help provide some guidance to resolve this issue? I tried both Mac M2 and Windows WSL (Ubuntu), both no luck.~~

Tried fixes in https://github.com/quarkusio/quarkus/pull/33681, it works fine now

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
viniciusfcf: Actually I don't know. Do you suggest to remove both lines?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
viniciusfcf: I have one newbie question: local.settins.json is enough to test locally (without host.json)?
I think the plugin should not overwrite any json files created by the user.

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
viniciusfcf: So, just to validate if I understand all:
1. Move getArtifact from AbstractFunctionMojo to Package
2. change nativeExecutable to nativeExecutablePath, and if nativeExecutablePath is defined, I will call copyBinaryToStageDirectory instead copyJarsToStageDirectory

Correct?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
SandraAhlgrimm: Yes, you can test locally without the host.json file.
local.settings.json is used for local development and will not be published to Azure. [On the other hand, host.json is used for configuring the Functions host process when you run projects locally and in Azure, for log level for instance]

https://learn.microsoft.com/en-us/azure/azure-functions/functions-develop-local

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
Flanker32: I prefer to keep line 357, so that we could track how many users use native executable, for custom handler args, I'm not sure whether we need track it, so may be it could be removed

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/PackageMojo.java
Flanker32: @viniciusfcf @SandraAhlgrimm Thank you all for your help. Just want to confirm whether customHandler configuration is required in host.json? I just tried to remove following configuration and func seems broken. 

If its required, do we need to add validation throw exception if user did not set it in host.json? Besides, I'm not sure whether we should provide default configuration as the following configuration should only works for quarkus, or is there any way we could detect whether the binary is a quarkus one or spring boot?

```
    "customHandler": {
        "description": {
            "defaultExecutablePath": "quarkus-azure-functions-http-1.0.0-SNAPSHOT",
            "workingDirectory": "",
            "arguments": [
                "-Dquarkus.http.port=${FUNCTIONS_CUSTOMHANDLER_PORT}"
            ]
        },
        "enableForwardingHttpRequest": true
    }
```

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
Flanker32: 1. I prefer to keep `getArtifact` as it is, and move binary related logic to `copyBinaryToStageDirectory` in package mojo.
2. Yes, I think `nativeExecutablePath` may be more flexible, and we no longer need to generate the path for executable

viniciusfcf: Hi @Flanker32 , I did not test all the changes yet.

edburns (dismissed):

SandraAhlgrimm:
> Converting to Draft. I'll seek guidance from @Flanker32 , but in general I am in favor of this feature.

Can we convert back to Open PR please 👍

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
```diff
@@ -239,6 +259,10 @@ protected File getArtifact() throws AzureToolkitRuntimeException {
         return result;
     }
 
+    protected boolean isNativeExecutable() {
+        return StringUtils.isEmpty(getNativeExecutablePath());
```
Flanker32: should be `isNotEmpty`?

azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
```diff
@@ -323,6 +347,7 @@ public Map<String, String> getTelemetryProperties() {
         result.put(DISABLE_APP_INSIGHTS_KEY, String.valueOf(isDisableAppInsights()));
         final boolean isDeployToFunctionSlot = getDeploymentSlotSetting() != null && StringUtils.isNotEmpty(getDeploymentSlotSetting().getName());
         result.put(FUNCTION_DEPLOY_TO_SLOT_KEY, String.valueOf(isDeployToFunctionSlot));
+        result.put(FUNCTION_NATIVE_EXECUTABLE_PATH, nativeExecutablePath);
```
Flanker32: please just track `isNativeExecutable` as path may include pii data

Flanker32:
@viniciusfcf @SandraAhlgrimm Thank you all for your help and efforts. Please see my comments in the PR.

Besides, for the configuration in host.json, just want to confirm whether customHandler configuration is required in host.json? If so, I think we should add validation in native binary cases and throw exception if use did not set it or the host.json is missing, as it seems that we could not provide a default configuration for this case (the default configuration in this PR seems only works for quarkus.

SandraAhlgrimm:
@agoncal for review

edburns (dismissed):

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/config/FunctionAppConfig.java
```diff
@@ -19,6 +19,8 @@ public class FunctionAppConfig extends AppServiceConfig {
     private boolean disableAppInsights;
     private String storageAccountName;
     private String storageAccountResourceGroup;
+    private String nativeExecutablePath;
```
Flanker32, line 22: Please remove these new parameters as `FunctionAppConfig` is used to create resource in Azure, which should not contains native binary configurations

PR #2307, opened 2023-06-12
[AB#2072686] Support toggle action visible status based on action place

Flanker32:
Support toggle action visible status based on action place, #2072686

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested


azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/action/Action.java
```diff
@@ -64,6 +64,7 @@
     private List<Function<D, String>> titleParamProviders = new ArrayList<>();
 
     private D source;
+    private String place;
```
wangmingliang-ms, line 67: place should not be a property of action. because action can be placed anywhere.

wangmingliang-ms: remove it in late PR, please.

PR #2309, opened 2023-06-15
Fix one deploy may be skipped as we only get the WebSiteBasic instance

Flanker32:
- Fix one deploy may be skipped as we only get the WebSiteBasic instance

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested


wangmingliang-ms (approved):
LGTM

PR #2322, opened 2023-07-10
Support create new cluster in spring apps maven plugin

Flanker32:
…

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested


azure-spring-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/springcloud/config/AppDeploymentRawConfig.java
```diff
@@ -19,4 +22,13 @@ public class AppDeploymentRawConfig {
     private String deploymentName;
     private String jvmOptions;
     private String runtimeVersion;
+
+    public void saveSpringCloudDeployment(@Nonnull final SpringCloudDeployment activeDeployment) {
```
wangmingliang-ms, line 26: rename to `from(...)`? see also `SpringCloudAppConfig/SpringCloudDeploymentConfig.from(xxx)`

azure-spring-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/springcloud/config/AppRawConfig.java
```diff
@@ -5,14 +5,32 @@
...
 public class AppRawConfig {
     private String subscriptionId;
     private String resourceGroup; // optional
     private String clusterName;
     private String appName;
     private String isPublic;
+    private String region;
+    private String sku;
+    private String environment;
+    private String environmentResourceGroup;
```
wangmingliang-ms: how about adding a `ClusterRawConfig`, as an alternative config to `clusterName`.

azure-toolkit-libs/azure-toolkit-springcloud-lib/src/main/java/com/microsoft/azure/toolkit/lib/springcloud/config/SpringCloudAppConfig.java
```diff
@@ -29,6 +31,10 @@
 public class SpringCloudAppConfig {
     private String subscriptionId;
     private String clusterName;
+    private String region;
+    private String sku;
+    private String environment;
+    private String environmentResourceGroup;
```
wangmingliang-ms: how about adding a `SpringCloudClusterConfig cluster`?

azure-spring-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/springcloud/ConfigMojo.java
```diff
@@ -154,8 +165,16 @@ protected void doExecute() throws AzureExecutionException {
             loginAzure();
             promptAndSelectSubscription();
 
-            selectAppCluster();
-            configCommon();
+            // prompt to select existing cluster or create a new one
+            useExistingCluster = this.wrapper.handleConfirm("Using existing cluster in Azure (Y/n):", true, true);
+            final SpringCloudCluster cluster = useExistingCluster ? selectAppCluster() : configCluster();
+            useExistingApp = !Objects.isNull(cluster) && !parentMode &&
+                this.wrapper.handleConfirm(String.format("Using existing app in cluster %s (Y/n):", cluster.getName()), true, true);
```
wangmingliang-ms: we will definitely create a new app if user choose to create a new cluster.

Flanker32: Yes, in this case `configureCluster()` will return null, so useExistingApp will be false in this case.

azure-spring-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/springcloud/config/AppDeploymentRawConfig.java
Flanker32: I prefer to keep current naming as this method is to save remote deployment configuration in an existing `AppDeploymentRawConfig` object while `from(...)` is more likely to create a new instance

Flanker32: -> merge

azure-spring-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/springcloud/ConfigMojo.java
```diff
@@ -116,6 +124,9 @@ public class ConfigMojo extends AbstractMojoBase {
     @Parameter(property = "advancedOptions")
     private Boolean advancedOptions;
 
+    private boolean useExistingCluster = false;
```
Flanker32, line 133: remove this from mojo prarameter lists

azure-spring-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/springcloud/ConfigMojo.java
```diff
@@ -171,12 +190,76 @@ protected void doExecute() throws AzureExecutionException {
         }
     }
 
+    private SpringCloudCluster configCluster() throws IOException, InvalidConfigurationException {
+        configureClusterName();
+        configureResourceGroup();
+        final Region region = configureRegion();
+        final Sku sku = configureSku(region);
+        if (sku.isConsumptionTier()) {
```
Flanker32: may comment it before consumption ga

PR #2356, opened 2023-08-25
Adding templates for Azure Functions Dapr extension in Java

ASHIQUEMD:
Adding Azure Functions Dapr extension Java template, it has three templates

1. DaprPublish
2. DaprServiceInvocationTrigger
3. DaprTopicTrigger

No

No

Any other comments?
-------------------
No

Has this been tested?
---------------------------
- [ ] Tested


azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/templates.json
```diff
@@ -451,6 +451,113 @@
...
...ts-path .\/components\/ -- mvn azure-functions:run\r\n    Invoke function app: dapr publish --pubsub messagebus --publish-app-id functionapp --topic A --data '{\\\"value\\\": { \\\"orderId\\\": \\\"42\\\" } }'\r\n    *\/\r\n    @FunctionName(\"DaprTopicTriggerFunc\")\r\n    public String run(\r\n        @DaprTopicTrigger(\r\n            pubSubName = \"messagebus\",\r\n            topic = \"X\")\r\n            String request,\r\n        @DaprStateOutput(\r\n            stateStore = \"statestore\",\r\n            key = \"product\")\r\n        OutputBinding<String> product,\r\n        final ExecutionContext context) throws JsonProcessingException {\r\n        context.getLogger().info(\"Java DaprTopic trigger with DaprState output binding function processed a request from the Dapr Runtime.\");\r\n\r\n        \/\/ Get the CloudEvent data from the request body as a JSON string\r\n        ObjectMapper objectMapper = new ObjectMapper();\r\n        JsonNode jsonNode = objectMapper.readTree(request).get(\"data\");\r\n        String serializedJson = objectMapper.writeValueAsString(jsonNode);\r\n\r\n        product.setValue(serializedJson);\r\n\r\n        return serializedJson;\r\n    }\r\n}",
```
Flanker32: please use `$packageName$` and `$className$` to replace the package name and class name

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/templates.json
```diff
@@ -451,6 +451,113 @@
...
+        "components\\cron_bindings.yaml": "apiVersion: dapr.io\/v1alpha1\r\nkind: Component\r\nmetadata:\r\n  name: scheduled\r\nspec:\r\n  type: bindings.cron\r\n  version: v1\r\n  metadata:\r\n  - name: schedule\r\n    value: \"@every 10s\" # valid cron schedule\r\n  - name: direction\r\n    value: \"input\"",
```
Flanker32: currently we do not support create multiple files with function java template

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/templates.json
Flanker32: It seems that we use `com.microsoft.azure.functions.dapr.*` in tempalte code, which should belong to `azure-functions-java-library-dapr`, we may also need to tell user introduce this dependency in pom.xml/build.gradle

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/bindings.json
```diff
@@ -526,6 +526,33 @@
...
         }
       ]
+    },
+    {
+      "type": "daprBindingTrigger",
+      "displayName": "$daprBindingTrigger_displayName",
+      "direction": "trigger",
+      "enabledInTryMode": true,
+      "settings": [
+
```
Flanker32, line 536: we may add some setting parameters here, so that `functions:add` could prompt users to set the required values. For example, `pubSubName` and `pubSubName`

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/templates.json
ASHIQUEMD: For Dapr we need this additional files, is there a way to support this?

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/templates.json
ASHIQUEMD: Where do we tell the user to add these dependencies? as a comment on top of function? 

Is it not possible to add this dependency in pom.xml from our side so that, it gets added automatically when user creates Azure functions with Dapr template?

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/bindings.json
ASHIQUEMD: These are hardcoded in the function itself because same name needs to be configured in component's yaml file. Thats the reason it is not added as a prompt.

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/templates.json
Flanker32: Currently `mvn azure-functions:add` will not modify users' pom configuration. However, this could be done with maven archetype, which we could customized dependency by parameters, please refer https://github.com/microsoft/azure-maven-archetypes/blob/develop/azure-functions-archetype/src/main/resources/archetype-resources/pom.xml#L26-L35

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/templates.json
Flanker32: We have added this item to our back log, as a workaround, you may also add the configuration files to maven archetype, and add them with groovy script here https://github.com/microsoft/azure-maven-archetypes/blob/develop/azure-functions-archetype/src/main/resources/META-INF/archetype-post-generate.groovy

ASHIQUEMD: I think it will become cumbersome if we add Dapr specific files with multiple if conditions. Because all three templates have different component files. Do we have any ETA for the backlog item?

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/templates.json
ASHIQUEMD: Here is the PR: https://github.com/microsoft/azure-maven-archetypes/pull/217/files

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/templates.json
Flanker32: @ASHIQUEMD Thanks for your efforts and sorry for the late response, we have prioritized this working item, which should be done in our October release.

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/resources/templates.json
Flanker32: @ASHIQUEMD Thanks for your efforts, I've added some comments to PR https://github.com/microsoft/azure-maven-archetypes/pull/217/files

wangmingliang-ms (approved):
LGTM

PR #2397, opened 2023-11-24
Change to use list() instead of send request agagin when load resource from Azure

Flanker32:
Change to use list() instead of send request agagin when load resource from Azure

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested


wangmingliang-ms (approved):
LGTM

PR #2413, opened 2023-12-21
refactor Runtime data structure of webapp/functionapp based on stack API.

wangmingliang-ms:
…

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested


azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/model/WebAppWindowsRuntime.java
```diff
@@ -0,0 +1,167 @@
...
+        this.javaVersionDisplayText = javaMinorVersion.displayText();
+    }
+
+    WebAppWindowsRuntime(final String containerUserText, final String javaVersionUserText) {
+        final String[] containerParts = containerUserText.split(" ");
+        final String[] javaParts = javaVersionUserText.split(" ");
+        this.deprecatedOrHidden = false;
+        this.containerName = containerParts[0];
+        this.containerVersionNumber = containerParts[1];
+        this.javaVersionNumber = javaParts[1];
```
Flanker32, line 117: will there be any NPE? as users did not use space but _

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/model/WebAppWindowsRuntime.java
```diff
@@ -0,0 +1,167 @@
...
+
+    private WebAppWindowsRuntime(@Nonnull WebAppMinorVersion webContainer, @Nonnull WebAppMinorVersion javaMinorVersion) {
+        final WindowsJavaContainerSettings containerSettings = webContainer.stackSettings().windowsContainerSettings();
+        final WebAppRuntimeSettings javaSettings = javaMinorVersion.stackSettings().windowsRuntimeSettings();
+        this.deprecatedOrHidden = BooleanUtils.isTrue(containerSettings.isHidden())
+            || BooleanUtils.isTrue(javaSettings.isHidden())
+            || BooleanUtils.isTrue(containerSettings.isDeprecated())
+            || BooleanUtils.isTrue(javaSettings.isDeprecated());
+        this.containerName = containerSettings.javaContainer().toUpperCase();
+        this.containerVersionNumber = containerSettings.javaContainerVersion().toUpperCase();
```
Flanker32: why we need upper case container version

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/model/WebAppRuntime.java
```diff
@@ -0,0 +1,139 @@
...
+
+    static List<WebAppRuntime> getAllRuntimes() {
+        return Stream.concat(
+                Stream.of(WebAppDockerRuntime.INSTANCE), Stream.concat(
+                    WebAppWindowsRuntime.getAllRuntimes().stream(),
+                    WebAppLinuxRuntime.getAllRuntimes().stream()))
+            .collect(Collectors.toList());
+    }
+
+    static List<PricingTier> getPricingTiers(OperatingSystem os, String containerUserText) {
```
Flanker32, line 109: why not use Runtime here instead of containerUserText

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/model/WebAppWindowsRuntime.java
wangmingliang-ms: > why we need upper case container version

@Flanker32  to keep consistency, versions are in uppercase for linux runtimes.   not sure if there are versions like `xxx_ZULU_xxx`

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/model/WebAppWindowsRuntime.java
wangmingliang-ms: This will never happen here, because we never **create** new Runtime instance from user input. we only try to find exsting one based on user input.

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/model/WebAppRuntime.java
wangmingliang-ms: because the caller doesn't specify `javaVersion`. seems it doesn't have it when calling this method.

azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/appservice/model/WebAppWindowsRuntime.java
```diff
@@ -133,7 +134,11 @@ public static WebAppWindowsRuntime fromContainerAndJavaVersion(final String cont
     }
 
     @Nullable
-    public static WebAppWindowsRuntime fromContainerAndJavaVersionUserText(final String containerUserText, final String javaVersionUserText) {
+    public static WebAppWindowsRuntime fromContainerAndJavaVersionUserText(final String containerUserText, String javaVersionUserText) {
+        if (StringUtils.isBlank(javaVersionUserText)) {
```
Flanker32, line 139: Not sure whether we should add this, because if user just want to deploy to an existing app, it may be a bit confused to show this message.

PR #2438, opened 2024-04-08
Azure SDK build tool maven plugin

srnagar:
This change adds Azure SDK build tool maven plugin that can be used by Azure SDK customers. The complete list of plugin features are in the README of the plugin library.

None

Sample logs
```
[ERROR]  - azure-security-keyvault-keys The azure-sdk-bom version is ignored and a dependency version is explicitly specified.
[ERROR]  - azure-monitor-opentelemetry-exporter The azure-sdk-bom version is ignored and a dependency version is explicitly specified.
[ERROR]  - azure-monitor-opentelemetry-exporter A beta dependency is used.
[ERROR]  - A direct dependency of this project relies on a `com.microsoft.*` library. Consider upgrading to the `com.azure.*` library listed below:
[ERROR]     - com.microsoft.azure:azure-cosmosdb:2.6.13 --> [com.azure:azure-cosmos:4.57.0]
[ERROR]     - com.microsoft.azure:azure-keyvault:1.2.6 --> [com.azure:azure-security-keyvault-keys:4.8.1, com.azure:azure-security-keyvault-certificates:4.6.1, com.azure:azure-security-keyvault-secrets:4.8.1]
[ERROR]  - A method annotated with Beta is called!
[ERROR]    - Method public com.azure.cosmos.ChangeFeedProcessorBuilder com.azure.cosmos.ChangeFeedProcessorBuilder.handleAllVersionsAndDeletesChanges(java.util.function.Consumer) is annotated with @Beta and called by public static void com.test.annotation.BetaApiTestApp.main(java.lang.String[])

```

Any other comments?
-------------------
None

Has this been tested?
---------------------------
- [x] Tested


azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/util/MavenUtils.java
```diff
@@ -0,0 +1,107 @@
...
+     * @param groupId The group id of the artifact.
+     * @param artifactId The artifact id of the artifact.
+     * @return The latest version or {@code null} if an error occurred while retrieving the latest
+     * version.
+     */
+    public static String getLatestArtifactVersion(String groupId, String artifactId) {
+        HttpURLConnection connection = null;
+        try {
+            groupId = groupId.replace(".", "/");
+            URL url = new URL("https://repo1.maven.org/maven2/" + groupId + "/" + artifactId + "/maven-metadata.xml");
```
wangmingliang-ms, line 57: we implemented similar features based on this file ([java-packages.csv](https://github.com/Azure/azure-sdk/blob/main/_data/releases/latest/java-packages.csv)) in our IntelliJ IDEA plugin, it provide more details for all libs. which may be useful if you want to implement more features in the future. and it can also help reduce http requests.

azure-sdk-build-tool/pom.xml
```diff
@@ -0,0 +1,319 @@
...
+    </developer>
+  </developers>
+
+  <issueManagement>
+    <system>GitHub</system>
+    <url>https://github.com/Azure/azure-sdk-for-java/issues</url>
+  </issueManagement>
+
+  <scm>
+    <url>https://github.com/Azure/azure-sdk-for-java</url>
```
wangmingliang-ms, line 59: are these links right as expect?

azure-sdk-build-tool/pom.xml
```diff
@@ -0,0 +1,319 @@
...
+    <dependency>
+      <groupId>net.oneandone.reflections8</groupId>
+      <artifactId>reflections8</artifactId>
+      <version>0.11.7</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core</artifactId>
+      <version>1.46.0</version>
```
wangmingliang-ms: why not use `azure-sdk-bom` as this plugin suggested?

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/mojo/AzureSdkMojo.java
```diff
@@ -0,0 +1,253 @@
...
+    private boolean validateNoBetaApiUsed;
+
+    @Parameter(property = "validateLatestBomVersionUsed", defaultValue = "true")
+    private boolean validateLatestBomVersionUsed;
+
+    @Parameter(property = "reportFile", defaultValue = "")
+    private String reportFile;
+
+    @Parameter(property = "sendToMicrosoft", defaultValue = "true")
+    private boolean sendToMicrosoft;
```
wangmingliang-ms, line 111: Would it be more compliant to ask the user if they would like to send data to Microsoft on the first run? Or at least tell the user how to shut down in log.

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/mojo/AzureSdkMojo.java
```diff
@@ -0,0 +1,253 @@
...
+    private static final String APP_INSIGHTS_INSTRUMENTATION_KEY = "1d377c0e-44f8-4d56-bee7-7f13a3fef594";
+    private static final String APP_INSIGHTS_ENDPOINT = "https://centralus-2.in.applicationinsights.azure.com/";
+    private static final String AZURE_SDK_BUILD_TOOL = "azure-sdk-build-tool";
+    private final ApplicationInsightsClient applicationInsightsClient;
+
+    @Parameter(defaultValue = "${project}", readonly = true, required = true)
+    private MavenProject project;
+
+    @Parameter(property = "validateAzureSdkBomUsed", defaultValue = "true")
+    private boolean validateAzureSdkBomUsed;
```
wangmingliang-ms, line 61: you can configure to [generate a good help mojo](https://maven.apache.org/plugin-tools/maven-plugin-plugin/examples/generate-help.html) if these parameter fields have well-writen javadocs.

wangmingliang-ms (commented):
I will continue to review later.

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/util/MavenUtils.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Gets the latest released version of the given artifact from Maven repository.
+     * @param groupId The group id of the artifact.
+     * @param artifactId The artifact id of the artifact.
+     * @return The latest version or {@code null} if an error occurred while retrieving the latest
+     * version.
+     */
+    public static String getLatestArtifactVersion(String groupId, String artifactId) {
```
wangmingliang-ms, line 53: this method can be possibly simplified if it's acceptable to introduce another dependency `jackson-dataformat-xml`.

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/util/MavenUtils.java
srnagar: Could you please share the place where these features are implemented? We are using Maven as that's the source of truth for customers as the java-packages.csv may get out of sync. But we can potentially use it to cache some details.

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/util/MavenUtils.java
srnagar: It's an additional dependency that may not be required for this simple usecase.

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/util/MavenUtils.java
wangmingliang-ms: @srnagar  here it is. https://github.com/microsoft/azure-tools-for-java/blob/68680d220a0199efc62e156a79821e1ef49d663c/PluginsAndFeatures/azure-toolkit-for-intellij/azure-sdk-reference-book/src/main/java/com/microsoft/azure/toolkit/intellij/azuresdk/service/AzureSdkLibraryService.java#L54

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/DependencyCheckerTool.java
```diff
@@ -0,0 +1,172 @@
...
+            MojoUtils.failOrWarn(AzureSdkMojo.getMojo()::isValidateAzureSdkBomUsed, BuildErrorCode.BOM_NOT_USED, MojoUtils.getString("missingBomDependency"));
+        }
+    }
+
+    private void checkForAzureSdkDependencyVersions() {
+        List<Dependency> dependencies = AzureSdkMojo.getMojo().getProject().getDependencies();
+        List<Dependency> dependenciesWithOverriddenVersions = dependencies.stream()
+                .filter(dependency -> dependency.getGroupId().equals("com.azure"))
+                .filter(dependency -> {
+                    InputLocation location = dependency.getLocation("version");
```
wangmingliang-ms, line 96: is `location` possibly null if `version` is not configured?

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/AzureDependencyMapping.java
```diff
@@ -0,0 +1,82 @@
...
+
+        // Service Bus
+        TRACK_ONE_REDIRECTS.put("azure-servicebus", Collections.singletonList("azure-messaging-servicebus"));
+
+        // Event Grid
+        TRACK_ONE_REDIRECTS.put("azure-eventgrid", Collections.singletonList("azure-messaging-eventgrid"));
+
+        // Log Analytics
+        TRACK_ONE_REDIRECTS.put("azure-loganalytics", Collections.singletonList("azure-monitor-query"));
+    }
```
wangmingliang-ms, line 53: `java-package.csv` has an official and more complete list for these.

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/DependencyCheckerTool.java
```diff
@@ -0,0 +1,172 @@
...
+            AzureSdkMojo.getMojo().getReport().setBomVersion(usedBomVersion);
+        } else {
+            MojoUtils.failOrWarn(AzureSdkMojo.getMojo()::isValidateAzureSdkBomUsed, BuildErrorCode.BOM_NOT_USED, MojoUtils.getString("missingBomDependency"));
+        }
+    }
+
+    private void checkForAzureSdkDependencyVersions() {
+        List<Dependency> dependencies = AzureSdkMojo.getMojo().getProject().getDependencies();
+        List<Dependency> dependenciesWithOverriddenVersions = dependencies.stream()
+                .filter(dependency -> dependency.getGroupId().equals("com.azure"))
```
wangmingliang-ms: use constant in the same way as that in `checkForAzureSdkTrackOneDependencies`?

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/DependencyCheckerTool.java
```diff
@@ -0,0 +1,172 @@
...
+            List<String> outdatedDependencyGavs = outdatedDirectDependencies
+                .stream()
+                .map(OutdatedDependency::getOutdatedDependency)
+                .collect(Collectors.toList());
+            MojoUtils.failOrWarn(AzureSdkMojo.getMojo()::isValidateNoDeprecatedMicrosoftLibraryUsed, BuildErrorCode.DEPRECATED_DEPENDENCY_USED, message.toString(), outdatedDependencyGavs);
+        }
+        if (!outdatedTransitiveDependencies.isEmpty()) {
+            // convert each track one dependency into actionable guidance
+            StringBuilder message = new StringBuilder(MojoUtils.getString("deprecatedIndirectDependency"));
+            for (OutdatedDependency outdatedDependency : outdatedDirectDependencies) {
```
wangmingliang-ms: `outdatedDirectDependencies` -> `outdatedTransitiveDependencies`?

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/AnnotationProcessingTool.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static com.microsoft.azure.sdk.build.tool.util.MojoUtils.failOrWarn;
+import static com.microsoft.azure.sdk.build.tool.util.MojoUtils.getString;
```
wangmingliang-ms: please remove unused imports.

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/AnnotationProcessingTool.java
```diff
@@ -0,0 +1,121 @@
...
+     */
+    public void run() {
+        LOGGER.info("Running Annotation Processing Tool");
+
+        // We build up a list of packages in the source of the user maven project, so that we only report on the
+        // usage of annotation methods from code within these packages
+        final Set<String> interestedPackages = new TreeSet<>(Comparator.comparingInt(String::length));
+        MojoUtils.getCompileSourceRoots().forEach(root -> buildPackageList(root, root, interestedPackages));
+
+        final ClassLoader classLoader = AnnotationUtils.getCompleteClassLoader(getAllPaths());
```
wangmingliang-ms: `getAllPaths` is expensive, and it's called multiple times in this method. please consider saving its return value in a local variable to improve performance.

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/util/AnnotationUtils.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+        return Optional.empty();
+    }
+
+    /**
+     * Returns a list of methods that call methods that are annotated with the given annotation.
+     * @param annotation The annotation on the method to look for.
+     * @param paths The paths to scan.
+     * @param interestedPackages The packages that this scan should be limited to.
+     * @param recursive If true, look for packages in the sub-directories of the given paths.
```
wangmingliang-ms: it seems it will also find **indirect** calls to the annotated methods when `recursive` is `true` according to the code. is my understanding right? please consider rename this param and rewrite the javadoc if my understanding is correct.

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/AnnotationProcessingTool.java
```diff
@@ -0,0 +1,121 @@
...
+
+        // Collect all calls to methods annotated with the Azure SDK @ServiceMethod annotation
+        Optional<Set<AnnotatedMethodCallerResult>> serviceMethodCallers = AnnotationUtils.getAnnotation("com.azure.core.annotation.ServiceMethod", classLoader)
+            .map(a -> AnnotationUtils.findCallsToAnnotatedMethod(a, getAllPaths(), interestedPackages, true));
```
wangmingliang-ms: `serviceMethodCallers` will also contains calls from a class (with a same package name in the `interestedPackages` but) in a dependency lib, but this is not expected, right? 
I think we are only interested in calls from the source code of current module, right?

wangmingliang-ms (changes requested):
1. build failed with the below error on my local machine (Apache Maven `3.9.6`) 
```
[ERROR] Failed to execute goal com.azure.tools:codesnippet-maven-plugin:1.0.0-beta.8:update-codesnippet (update-codesnippets) on project azure-sdk-build-tool: The plugin com.azure.tools:codesnippet-maven-plugin:1.0.0-beta.8 requires Maven version 3.6.3 -> [Help 1]
```
2. we prefer small PR or PR with small commits (<200 modified lines)

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/AzureDependencyMapping.java
srnagar: The data in java-package.csv is not accurate for finding replacements. So, I will skip this for now and follow up with maintainers of java-package.csv to get this fixed.

azure-sdk-build-tool/src/main/java/com/microsoft/azure/sdk/build/tool/AnnotationProcessingTool.java
srnagar: As discussed offline, this is not a concern if we find a service method call in a package that's not in the current package. We still want to record the service method if the source makes a call transitively through another dependency.

wangmingliang-ms (approved):
LGTM

PR #2513, opened 2024-11-21
Add commom check logic for source code and artifact in toolkit-lib

RuoyuWang-MS:
…

1. Fix jenny's comments, print more infos when deploying successfully.
2. Add common check logic for source code and artifact in toolkit-lib. Only check if the jar is executable because we may need to support quarkus later. I test this and it works fine with quarkus uber jar.
3. Fix a small bug when deploying failed. 

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested

### DIFF
--- a/azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/utils/Utils.java
+++ b/azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/utils/Utils.java
@@ -27,6 +27,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.BufferedOutputStream;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -57,6 +58,7 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
+import java.util.jar.JarInputStream;
 import java.util.jar.Manifest;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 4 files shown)


azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/utils/Utils.java
```diff
@@ -369,4 +371,53 @@ public FileVisitResult visitFileFailed(Path file, IOException exc) {
...
         return tarFilePath;
     }
+
+    public static boolean isSpringBootProject(Path projectPath) {
+        try {
+            // Check if `pom.xml` exists and contains Spring Boot references
+            Path pomPath = projectPath.resolve("pom.xml");
+            if (Files.exists(pomPath) &&
+                (containsText(pomPath, "org.springframework.boot") ||
+                    containsText(pomPath, "spring-boot-maven-plugin"))) {
```
allxiao, line 381: Can we use more accurate detection? For example check for XML elements rather than using raw text match?

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/utils/Utils.java
```diff
@@ -369,4 +371,53 @@ public FileVisitResult visitFileFailed(Path file, IOException exc) {
...
+                return true;
+            }
+
+            // Check if `build.gradle` exists and contains Spring Boot references
+            Path gradlePath = projectPath.resolve("build.gradle");
+            if (Files.exists(gradlePath) &&
+                (containsText(gradlePath, "org.springframework.boot") ||
+                    containsText(gradlePath, "spring-boot-starter"))) {
```
allxiao, line 389: A probably better way would be checking for plugins `org.springframwork.boot` and `io.spring.dependency-management`.

The same question, should we check for more context rather than short, raw text match?

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/utils/Utils.java
```diff
@@ -369,4 +371,53 @@ public FileVisitResult visitFileFailed(Path file, IOException exc) {
...
+                return true;
+            }
+
+            // Check for Spring Boot configuration files
+            Path resourcesPath = projectPath.resolve("src/main/resources");
+            if (Files.exists(resourcesPath)) {
+                if (Files.exists(resourcesPath.resolve("application.properties")) ||
+                    Files.exists(resourcesPath.resolve("application.yml"))) {
```
allxiao, line 397: `application.yaml` is also a possible case. However, is Spring Boot the only case that uses `application.xxx` as configuration file?

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/utils/Utils.java
```diff
@@ -369,4 +371,53 @@ public FileVisitResult visitFileFailed(Path file, IOException exc) {
...
+                    return true;
+                }
+            }
+        } catch (IOException e) {
```
allxiao, line 401: In which case will we meet the exception? Given that we had some check on existence, I think it would be better we output some error in this case to notify the customer about the exception, rather than fail silently and treat it as non-spring-boot app.

azure-toolkit-libs/azure-toolkit-common-lib/src/main/java/com/microsoft/azure/toolkit/lib/common/utils/Utils.java
allxiao: Do we need to check for Gradle case? Given that we are building a Maven plugin?

PR #2517, opened 2024-12-09
Support managed identity to pull image from ACR

RuoyuWang-MS:
…

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested

### DIFF
--- a/azure-container-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/containerapps/AbstractMojoBase.java
+++ b/azure-container-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/containerapps/AbstractMojoBase.java
@@ -62,8 +62,8 @@ public abstract class AbstractMojoBase extends AbstractAzureMojo {
     protected ContainerRegistryConfig registry;
 
     @Getter
-    @Parameter(property = "registryIdentity")
-    protected String registryIdentity;
+    @Parameter(property = "identity")
+    protected String identity;
 
     @Getter
     @Parameter(property = "containers")
--- a/azure-container-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/containerapps/AbstractMojoBase.java
+++ b/azure-container-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/containerapps/AbstractMojoBase.java
@@ -62,6 +62,10 @@ public abstract class AbstractMojoBase extends AbstractAzureMojo {
     protected ContainerRegistryConfig registry;
 
     @Getter
+    @Parameter(property = "registryIdentity")
+    protected String registryIdentity;
+
+    @Getter
     @Parameter(property = "containers")
     protected List<AppContainerMavenConfig> containers;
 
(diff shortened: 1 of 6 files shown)


azure-toolkit-libs/azure-toolkit-containerapps-lib/src/main/java/com/microsoft/azure/toolkit/lib/containerapps/containerapp/ContainerAppDraft.java
```diff
@@ -414,15 +433,76 @@ private static Secret getSecret(final ImageConfig config) {
...
+                return new RegistryCredentials().withServer(registry.getLoginServerUrl()).withIdentity("system");
+            } else {
+                return new RegistryCredentials().withServer(registry.getLoginServerUrl()).withIdentity(config.registryIdentity);
+            }
         }
         return null;
     }
 
+    @Nullable
+    private ManagedServiceIdentity getManagedServiceIdentity(ImageConfig imageConfig) {
```
allxiao: This is not a pure get method. It contains permission granting. Suggest to clarify in the method name.

allxiao: And it seems only user assigned identity will be returned, which should also clarify in the method name.

azure-container-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/containerapps/AbstractMojoBase.java
```diff
@@ -61,6 +61,10 @@ public abstract class AbstractMojoBase extends AbstractAzureMojo {
     @Parameter(property = "registry")
     protected ContainerRegistryConfig registry;
 
+    @Getter
+    @Parameter(property = "registryIdentity")
```
allxiao: I wonder whether we need to explicitly say `registry` in this configuration. There are two concerns
1. if this is purely registry related setting, it may be better to place `registry` and `registryIdentity` in a group, which results in `{registry: {server: xxx, identity: xxx}}` configuration structure. However, this is likely to be a breaking change
2. identity is an ACA resource concept, and this identity might possibly be used in other scenarios (although it may be true that in Maven plugin context we only cares about AcrPull). Placing `registry` in front may cause confusion.

azure-toolkit-libs/azure-toolkit-containerapps-lib/src/main/java/com/microsoft/azure/toolkit/lib/containerapps/containerapp/ContainerAppDraft.java
```diff
@@ -414,15 +433,76 @@ private static Secret getSecret(final ImageConfig config) {
...
+            grantACRPullPermissionToIdentity(imageConfig, principalId);
+            return null;
+        }
+        try {
+            Identity identity = Azure.az(AzureManagedIdentity.class).getById(imageConfig.getRegistryIdentity());
+            grantACRPullPermissionToIdentity(imageConfig, identity.getPrincipalId());
+            final String identityJson = String.format("{\"principalId\" : \"%s\", \"clientId\" : \"%s\"}", identity.getPrincipalId(), identity.getClientId());
+            final SerializerAdapter adapter = SerializerFactory.createDefaultManagementSerializerAdapter();
+            // todo: sync with sdk team to find a better way to create user identity object
+            final UserAssignedIdentity userAssignedIdentity = adapter.deserialize(identityJson, UserAssignedIdentity.class, SerializerEncoding.JSON);
```
allxiao: Hmm, not related to this PR, but this is a wired design. It means there is no easy way to create a resource with user assigned managed identity.

The SDK library of appcontainers starting from 1.0.0 has a static method `fromJson`, but that doesn't help either.

allxiao (approved):

PR #2521, opened 2025-01-02
init config goal for aca maven plugin

RuoyuWang-MS:
…

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [ ] Tested


azure-container-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/containerapps/config/AppContainerMavenConfig.java
```diff
@@ -15,6 +18,9 @@
 
 @Getter
 @Setter
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
```
allxiao: Just `@Data`?

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
allxiao, line 1: We should not use `aca`, `asa` outside of Microsoft. Consider changing to `container-apps` or `containerapps`. Same for `asa` in file names and folder names.

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
```diff
@@ -0,0 +1,169 @@
+---
+id:               configure-app-env-name
+promote:          "Input the Azure Container App Environment name (***${default_val}***):"
```
allxiao: For product name, use Container App***s***

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
```diff
@@ -0,0 +1,169 @@
...
+default_val:      'env-${project.getArtifactId().replaceAll("[^-a-z0-9]", "").toLowerCase()}'
```
allxiao: Is it allowed to use two or more consecutive `-` in the env name? Please also check the naming restrictions for the other resources.

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
```diff
@@ -0,0 +1,169 @@
...
+required:         true
+auto_select:      true
+message:
+  empty_options:  "Cannot find any subscriptions."
+  auto_select:    "There is only one subscription (***${ subscriptions[0].displayName() }***) in your account, will use it automatically."
+  select_none:    "You have not selected any subscriptions, 'config' will terminate."
+default_index:    0
+
+---
+id:               select-ACA-env
```
allxiao: ACA -> container-apps

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
```diff
@@ -0,0 +1,169 @@
...
+promote:
+  header:       "Available Azure Container App Environments:"
```
allxiao: Apps

As previously learned from the content writer, the naming convention is:

* For the product name, use `Azure Container Apps` or `Container Apps`
* To reference a specific instance, use `container app`     (singular)

Please check and update all the namings in this PR.

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
```diff
@@ -0,0 +1,169 @@
...
+  auto_select:    "There is only one Azure Container App Environment (***${ environments[0].name() }***) in current subscription, will use it automatically."
+  select_none:    "You have not selected any Azure Container App Environment, 'config' will terminate."
+required:         true
+auto_select:      true
+default_index:    0
+
+---
+id:               select-app
+promote:
+  header:       "Available Azure Container Apps:"
```
allxiao: It seems the indentation is not fully aligned in this file. Suggest to re-indent some lines. For example, the string in this two lines need two more spaces indentation.

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
```diff
@@ -0,0 +1,169 @@
...
+promote:          "Input the app name (***${default_val}***):"
+resource:         App
+property:         appName
+default_val:          '${project.getArtifactId().replaceAll("[^-a-z0-9]", "").toLowerCase()}'
+expression:       true
+required:         true
+
+---
+id:               configure-deployment-type
+promote:          "Input deployment type for apps (***${schema.default}***):"
```
allxiao: Should this be an option selection? In the config step, I don't know what to input.

azure-maven-plugin-lib/src/main/resources/schema/com/microsoft/azure/toolkit/maven/asa/Cluster.json
```diff
@@ -0,0 +1,22 @@
+{
+	"$schema": "http://json-schema.org/draft-07/schema#",
```
allxiao: Please unify the indentation style and align with existing schema indentations. I see 2 spaces, 4 spaces, 8 spaces in this PR.

azure-spring-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/springcloud/ConfigMojo.java
```diff
@@ -157,7 +157,7 @@ protected void doExecute() throws AzureExecutionException {
         }
         final ExpressionEvaluator expressionEvaluator = new PluginParameterExpressionEvaluator(session, mojoExecution);
         try {
-            this.wrapper = new ConfigurationPrompter(expressionEvaluator);
+            this.wrapper = new ConfigurationPrompter(expressionEvaluator, "asa");
```
allxiao: avoid using `asa` outside.

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
```diff
@@ -0,0 +1,169 @@
...
+required:         true
+auto_select:      true
+default_index:    0
+
+---
+id:               select-registry
+promote:
+  header:       "Available Azure Container Registries:"
+  many:         "Select Azure Container Registry for deployment:"
+message:
```
allxiao: What's the internal logic here to select the container registry? I followed the guide to use existing registry, but it then pop up and ask me to input the registry name, instead of select registries.

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
```diff
@@ -0,0 +1,169 @@
...
+
+---
+id:               configure-deployment-type
+promote:          "Input deployment type for apps (***${schema.default}***):"
+resource:         App
+property:         deploymentType
+
+---
+id:               configure-code-dir
+promote:          "Input directory for source code (***${schema.default}***):"
```
allxiao: It seems the default value is pointing to the Maven plugin source folder instead of the customer's project root. For example, I see `/home/menxiao/Code/azure-maven-plugins/azure-maven-plugin-lib` default value.

I'm wondering if we need to provide this in the config step. I think just take the default source root would be sufficient?

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
```diff
@@ -0,0 +1,169 @@
...
+property:         image
+
+---
+id:               configure-registry-name
+promote:          "Input registry name:"
+resource:         App
+property:         registryName
+
+---
+id:               configure-min-replicas
```
allxiao: By default, `Configure Scale (Y/n)` has default value to Yes. I wonder if we should change the default to No, and make the config process easier to follow.

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
```diff
@@ -0,0 +1,169 @@
...
+promote:          "Input min replicas for apps [***${schema.minimum}***-***${schema.maximum}***] (***${schema.default}***):"
+resource:         App
+property:         minReplicas
+
+---
+id:               configure-max-replicas
+promote:          "Input max replicas for apps of standard consumption plan [***${schema.minimum}***-***${schema.maximum}***] (***${schema.default}***):"
```
allxiao: Max replica has default value to `10`, which is likely to be too large. I suggest we start with 2, so the default scale range is 0 to 2.

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
allxiao: The config process use an absolute path as default value, which gives the impression to customer that we need to use absolute path. However, it doesn't make sense to specify an absolute path at all. E.g., I specified `/home/menxiao/tmp/demo`. However, this is likely to change when this is pushed to upstream, and then cloned by other people to other places.

I think we should skip this configuration. The directory should be default to project root. And if the customer needs to update, they can refer to our documentation and update in the pom file.

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
RuoyuWang-MS: Correct the default value as true to use existing ACR by default.

azure-maven-plugin-lib/src/main/resources/MessageTemplates-aca.yaml
RuoyuWang-MS: The ACA quickstart app use 0-10 as the default min/max replicas. Keep the same logic.

azure-container-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/containerapps/ConfigMojo.java
```diff
@@ -0,0 +1,432 @@
...
+
+        }
+    }
+
+    private ContainerAppsEnvironment configAppEnv() throws IOException, InvalidConfigurationException {
+        configureAppEnvName();
+        configureResourceGroup();
+        configureRegion();
+
+        return null;
```
Flanker32: Why always returns null here? shall we just change to void if the result is not necessary?

azure-container-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/containerapps/ConfigMojo.java
```diff
@@ -0,0 +1,432 @@
...
+            log.info(String.format("Using Azure Container Registry: %s", TextUtils.blue(targetRegistry.getName())));
+        }
+        this.registry = new ContainerRegistryConfig();
+        this.registry.setRegistryName(targetRegistry.getName());
+        this.registry.setResourceGroup(targetRegistry.getResourceGroup().getName());
+    }
+
+    private void configureIdentity() throws IOException, InvalidConfigurationException {
+        final boolean useExistingIdentity = this.wrapper.handleConfirm("Use existing User Assigned Identity (y/N):", false, true);
+        if (useExistingIdentity) {
```
Flanker32, line 203: no parameter is needed if create a new identity? Do we just uses system assigned managed identity in this case? Shall we show a info here?

RuoyuWang-MS: Here we use the managed identity for image pulling. But we cannot assign ACR_PULL role for the system assigned identity before it is created. So we suggest use user assigned identtiy (Here we prompt users to select a existing user assigned identity) or use the system assigned identity only if it is already created.

azure-container-apps-maven-plugin/src/main/java/com/microsoft/azure/maven/containerapps/parser/ConfigParser.java
```diff
@@ -65,7 +68,22 @@ public ContainerAppDraft.ImageConfig getImageConfigFromContainers(ContainerAppCo
         }
         if (containers.get(0).getDeploymentType() == DeploymentType.CODE || containers.get(0).getDeploymentType() == DeploymentType.ARTIFACT) {
             ContainerAppDraft.BuildImageConfig buildImageConfig = new ContainerAppDraft.BuildImageConfig();
-            buildImageConfig.setSource(Paths.get(containers.get(0).getDirectory()));
+            Path source = null;
+            if (containers.get(0).getDirectory() == null) {
```
allxiao: We have multiple places using `containers.get(0)`, better to pull it to a variable.

allxiao (dismissed):

Flanker32 (approved):
LGTM

PR #2532, opened 2025-03-28
Sdk types

ahmedmuhsin:

If a user uses a supported SDK Type in their function, the function.json should include a supportedDefferedBinding flag with a value of true in the properties dict.

For example, the following function uses BlobClient:

```java
package com.function;

import com.microsoft.azure.functions.annotation.*;
import com.microsoft.azure.functions.*;

import java.io.ByteArrayOutputStream;
import java.io.InputStream;
import java.nio.charset.StandardCharsets;

import com.azure.storage.blob.BlobClient;

/**
 * Azure Functions with Azure Blob trigger.
 */
public class BlobTriggerJava1 {
    /**
     * This function will be invoked when a new or updated blob is detected at the specified path. The blob contents are provided as input to this function.
     */
    @FunctionName("BlobTriggerJava1")
    @StorageAccount("AzureWebJobsStorage")
    public void run(
        @BlobTrigger(name = "content", path = "testcontainer/{name}", dataType = "binary", connection = "AzureWebJobsStorage") BlobClient blobClient,
        @BindingName("name") String name,
        final ExecutionContext context
    ) {
        // Read the blob content directly to a String
        try (ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
             InputStream inputStream = blobClient.openInputStream()) {
            
            System.out.println("Downloading blob content...");

            // Read from the InputStream into the ByteArrayOutputStream
            byte[] buffer = new byte[8192]; // Buffer size
            int bytesRead;
            while ((bytesRead = inputStream.read(buffer)) != -1) {
                outputStream.write(buffer, 0, bytesRead);
            }

            // Convert the output stream to a String
            String blobContent = outputStream.toString(StandardCharsets.UTF_8);

            // Print the blob content
            System.out.println("Blob content:");
            System.out.println(blobContent);

        } catch (Exception e) {
            System.err.println("Error reading blob: " + e.getMessage());
        }
    }
}
```

It's function.json should look like this:
```json
{
  "scriptFile" : "../blobtrigger-1.0-SNAPSHOT.jar",
  "entryPoint" : "com.function.BlobTriggerJava1.run",
  "bindings" : [ {
    "type" : "blobTrigger",
    "direction" : "in",
    "name" : "content",
    "path" : "testcontainer/{name}",
    "dataType" : "binary",
    "connection" : "AzureWebJobsStorage",
    "properties": {
      "supportsDeferredBinding": "true"
    }
  } ]
}
```

To facilitate this a new sdktypes library has been added to java-additions: https://github.com/Azure/azure-functions-java-additions/pull/38

Any other comments?
-------------------
…

Has this been tested?
---------------------------
- [x] Tested

### DIFF
--- a/azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/legacy/function/handlers/AnnotationHandlerImpl.java
+++ b/azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/legacy/function/handlers/AnnotationHandlerImpl.java
@@ -131,7 +131,7 @@ public class AnnotationHandlerImpl implements AnnotationHandler {
         SdkParameterAnalyzer analyzer = new SdkParameterAnalyzer();
         SdkParameterAnalysisResult result = analyzer.analyze(method);
         Set<Parameter> sdkTypeParams = new HashSet<>();
-        for (final SdkTypeMetaData sdkTypeMetaData : result.getSdkTypes()) {
+        for (final SdkTypeMetaData sdkTypeMetaData : result.getSdkTypesMetaData()) {
             sdkTypeParams.add(sdkTypeMetaData.getParam());
         }
 
--- a/azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/legacy/function/handlers/AnnotationHandlerImpl.java
+++ b/azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/legacy/function/handlers/AnnotationHandlerImpl.java
@@ -105,7 +105,10 @@ public class AnnotationHandlerImpl implements AnnotationHandler {
         SdkParameterAnalyzer analyzer = new SdkParameterAnalyzer();
         SdkParameterAnalysisResult result = analyzer.analyze(method);
         if(result.hasAnySdkTypes()){
-            bindings.forEach(binding -> binding.setAttribute("properties", "supportsDefferedBings: true"));
+            Map<String, String> props = new HashMap<>();
+            props.put("supportsDeferredBinding", "true");
+            binding.setAttribute("properties", props);
+            bindings.forEach(binding -> binding.setAttribute("properties", props));
         }
 
         config.setRetry(getRetryConfigurationFromMethod(method));
--- a/azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/legacy/function/handlers/AnnotationHandlerImpl.java
+++ b/azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/legacy/function/handlers/AnnotationHandlerImpl.java
@@ -136,13 +136,14 @@ public class AnnotationHandlerImpl implements AnnotationHandler {
         }
 
         for (final Parameter param : method.getParameters()) {
-            bindings.addAll(parseAnnotations(param::getAnnotations, this::parseParameterAnnotation));
-
+            List<Binding> paramBindings = parseAnnotations(param::getAnnotations, this::parseParameterAnnotation);
             if (sdkTypeParams.contains(param)) {
                 Map<String, String> props = new HashMap<>();
                 props.put("supportsDeferredBinding", "true");
-                bindings.forEach(binding -> binding.setAttribute("properties", props));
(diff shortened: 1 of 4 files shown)


azure-toolkit-libs/azure-toolkit-appservice-lib/src/main/java/com/microsoft/azure/toolkit/lib/legacy/function/handlers/AnnotationHandlerImpl.java
```diff
@@ -123,8 +128,21 @@ private Retry getRetryConfigurationFromMethod(Method method) throws AzureExecuti
...
+        SdkParameterAnalysisResult result = analyzer.analyze(method);
+        Set<Parameter> sdkTypeParams = new HashSet<>();
+        for (final SdkTypeMetaData sdkTypeMetaData : result.getSdkTypesMetaData()) {
+            sdkTypeParams.add(sdkTypeMetaData.getParam());
+        }
+
         for (final Parameter param : method.getParameters()) {
             bindings.addAll(parseAnnotations(param::getAnnotations, this::parseParameterAnnotation));
+
+            if (sdkTypeParams.contains(param)) {
```
Flanker32, line 140: Should we add `supportsDeferredBinding` to all bindings or binding related with param?

ahmedmuhsin: The flag should only be added to the binding related to the param that is a supported SDK type. 

For example, this function 

```java 
    @FunctionName("BlobTriggerUsingBlobClientToBlobTest")
    @StorageAccount("AzureWebJobsStorage")
    public void BlobTriggerToBlobTest_BlobClient(
            @BlobTrigger(name = "triggerBlob", path = "test-triggerinput-blobclient/{name}", dataType = "binary") BlobClient triggerBlobClient,
            @BindingName("name") String fileName,
            @BlobOutput(name = "outputBlob", path = "test-output-java-new/file.txt", dataType = "binary") OutputBinding<byte[]> outputBlob,
            final ExecutionContext context
    ) {
        context.getLogger().info("BlobTriggerUsingBlobClient triggered for blob: " + fileName);

        // Download the blob content
        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
        triggerBlobClient.downloadStream(outputStream);

        // Set the downloaded content as output
        outputBlob.setValue(outputStream.toByteArray());
        context.getLogger().info("Blob content copied successfully from trigger blob to output blob.");
    }
```

produces this function.json
```
{
  "scriptFile" : "../blobtrigger-1.0-SNAPSHOT.jar",
  "entryPoint" : "com.function.BlobTriggerJava1.BlobTriggerToBlobTest_BlobClient",
  "bindings" : [ {
    "type" : "blobTrigger",
    "direction" : "in",
    "name" : "triggerBlob",
    "path" : "test-triggerinput-blobclient/{name}",
    "dataType" : "binary",
    "connection" : "AzureWebJobsStorage",
    "properties" : {
      "supportsDeferredBinding" : "true"
    }
  }, {
    "type" : "blob",
    "direction" : "out",
    "name" : "outputBlob",
    "path" : "test-output-java-new/{name}",
    "dataType" : "binary",
    "connection" : "AzureWebJobsStorage"
  } ]
}
```

Flanker32: Got it, if so, I think we should move below logic to `parseAnnotations`, if we put it here it may also add 
`supportsDeferredBinding` to previous bindings
```
                Map<String, String> props = new HashMap<>();
                props.put("supportsDeferredBinding", "true");
                bindings.forEach(binding -> binding.setAttribute("properties", props));
```

ahmedmuhsin: Ah, I see the issue now. The bindings that I am modifying are an accumulation of all of the bindings for the method and not just that one param.

I have updated the logic so the flag is only added to the bindings for the param before they are added to the function's bindings. 

I did not want to move the logic to `parseAnnotations` since that would require me to modify its signature to pass in the param's analysis result (bool). Also, I think keeping all of the logic in one place makes it more readable.

Flanker32 (dismissed):
LGTM

ahmedmuhsin:
Thanks @Flanker32 .

Can you also run the CI pipeline and merge the PR? I don't have access to do so.

Flanker32 (approved):
LGTM

PR #2563, opened 2026-07-16
[#2562] Fix app insights wiring with connection string

### LINKED ISSUE #2562: Azure Functions Maven plugin auto-configures deprecated App Insights instrumentation key setup
Hi there,

It seems there is an issue with the Azure Functions Maven plugin App Insights integration, where the auto-creation/linking of Application Insights ends up wiring `APPINSIGHTS_INSTRUMENTATIONKEY`. That creates a deprecated setup by default.

Deprecation notice from Microsoft:
https://azure.microsoft.com/en-gb/updates?id=technical-support-for-instrumentation-key-based-global-ingestion-in-application-insights-will-end-on-31-march-2025

What we expected instead:

1. If the plugin auto-creates (or auto-links) an Application Insights instance, it should wire the `APPLICATIONINSIGHTS_CONNECTION_STRING` on the Function App.
2. The plugin should expose a dedicated connection string property in configuration (currently we only see `appInsightsKey`, `appInsightsInstance`, `disableAppInsights`).

Right now we need a post-deploy workaround to:

• fetch the App Insights connection string,
• set `APPLICATIONINSIGHTS_CONNECTION_STRING`,
• remove `APPINSIGHTS_INSTRUMENTATIONKEY`.

I think the correct behavior should be to adapt the plugin to wire the connection string directly, rather than the deprecated instrumentation key setting.

Would a pull request be welcome?
Br, Matthias

rascmatt:

This PR adds support for Application Insights Connection Strings in the Azure Functions Maven plugin. It addresses the fact that Instrumentation Keys have been deprecated a while ago: https://azure.microsoft.com/en-gb/updates?id=technical-support-for-instrumentation-key-based-global-ingestion-in-application-insights-will-end-on-31-march-2025

Main changes:
- Added a new config option: `appInsightsConnectionString`.
- Wired it through parsing/config (`AbstractFunctionMojo` -> `ConfigParser` -> `FunctionAppConfig` / `ApplicationInsightsConfig`).
- Updated Function App create/update/slot flows to handle `APPLICATIONINSIGHTS_CONNECTION_STRING`.
- Kept `appInsightsKey` support for backward compatibility, but prefer connection string when available.
- Added validation so users cannot set both `appInsightsKey` and `appInsightsConnectionString` explicitly.
- Improved redeploy behavior: if neither key nor connection string is explicitly configured, only the insights instance, deployment can resolve App Insights and migrate to connection-string based wiring.

Closes #2562

no

Any other comments?
-------------------
Goal here was to improve default behavior for new and redeployed apps without breaking existing users who still rely on `appInsightsKey`.

Has this been tested?
---------------------------
- [x] Tests in this PR
- [x] Actual deployment to Azure Cloud

### DIFF
--- a/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
+++ b/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/AbstractFunctionMojo.java
@@ -143,12 +143,23 @@ public abstract class AbstractFunctionMojo extends AbstractAppServiceMojo {
     protected String appInsightsInstance;
 
     /**
-     * Instrumentation key of the application insights instance
+     * Instrumentation key of the application insights instance.
+     * Will be skipped if `appInsightsConnectionString` is specified
      *
      * @since 1.6.0
      */
+    @Deprecated
     @Parameter(property = "functions.appInsightsKey")
     protected String appInsightsKey;
+
+    /**
+     * Connection string of the application insights instance.
+     *
+     * @since 1.46.0
+     */
+    @Getter
+    @Parameter(property = "functions.appInsightsConnectionString")
+    protected String appInsightsConnectionString;
 
     /**
      * Boolean flag to monitor the Function App with application insights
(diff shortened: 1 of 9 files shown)


rascmatt:
@microsoft-github-policy-service agree

